PR #2, opened 2020-01-09
implement base58

AharonNavi:
Implement next features:

- base58 for serrialization/desserialization functionality
- base58 tests

### DIFF
--- a/include/veriblock/base58.h
+++ b/include/veriblock/base58.h
@@ -1,13 +1,12 @@
 #ifndef _BASE58_H
 #define _BASE58_H
 
-#include <vector>
 #include <string>
+#include <vector>
 
-namespace Veriblock
-{
-    std::string          base58_encode(const void* buf, size_t size);
-    std::vector<uint8_t> base58_decode(const std::string& str);
-}
+namespace Veriblock {
+std::string base58_encode(const void *buf, size_t size);
+std::vector<uint8_t> base58_decode(const std::string &str);
+} // namespace Veriblock
 
-#endif 
+#endif
--- a/include/veriblock/base58.h
+++ b/include/veriblock/base58.h
@@ -1,5 +1,5 @@
-#ifndef _VBCRYPTO
-#define _VBCRYPTO
+#ifndef _BASE58_H
+#define _BASE58_H
 
 #include <vector>
 #include <string>
--- a/include/veriblock/base58.h
+++ b/include/veriblock/base58.h
@@ -1,0 +1,13 @@
+#ifndef _VBCRYPTO
+#define _VBCRYPTO
+
+#include <vector>
+#include <string>
+
+namespace Veriblock
+{
+    std::string          base58_encode(const void* buf, size_t size);
+    std::vector<uint8_t> base58_decode(const std::string& str);
+}
+
+#endif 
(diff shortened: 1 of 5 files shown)


include/veriblock/base58.h
```diff
@@ -0,0 +1,13 @@
+#ifndef _VBCRYPTO
+#define _VBCRYPTO
+
+#include <vector>
+#include <string>
+
+namespace Veriblock
+{
+    std::string          base58_encode(const void* buf, size_t size);
+    std::vector<uint8_t> base58_decode(const std::string& str);
```
Warchant: use clang-format for automatic formatting

include/veriblock/base58.h
```diff
@@ -0,0 +1,13 @@
+#ifndef _VBCRYPTO
+#define _VBCRYPTO
```
Warchant: did you write this include guard yourself?
if so, please use filename, such as `BASE58_HPP_` or similar

src/crypto/base58.cpp
```diff
@@ -0,0 +1,184 @@
...
+//==========================================================
+std::string base58_encode(const void* buf, size_t size)
+{
+    std::string result; 
+    std::vector<uint8_t> vch(size);
+    if (vch.empty()) {
+        return result;
+    }
+    memcpy(vch.data(), buf, size);
+    result = EncodeBase58(vch);
```
Warchant: {ptr, size} arguments are used when you do not need to copy buffer to any internal buffer, and can use {ptr,size} to read the data. 
Here is no point in copying data to vector, since you can easily call 
```
uint8_t* input = (uint8_t*) buf;
EncodeBase58(input, input + size);
```
and you can remove `EncodeBase58(vector)`

test/base58_test.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include <veriblock/base58.h>
+
+
+TEST(base58, Encode)
+{
+     //  test data
+     std::vector<uint8_t> inputA = { 27,53,84,6,78,51,29,38,57,2,0,95,23,89,4,6,7,92,35,61,30,95,12,96,29 };
+     std::vector<uint8_t> inputB = { 2,9,5,7,9,3,4,5,7,3,8,9,4,6,7,3,7,8,9,6,7,9,5,7,2,9,8,5,2,9,8,5,7,2 };
+     std::vector<uint8_t> inputC = { 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0 };
+     std::vector<uint8_t> inputD = { };
```
Warchant: this looks like separate test cases.
much better would be to create a test case
```
struct TestCase {
  std::vector<uint8_t> input;
  std::string output;
};
```

And create 2 value-parametrized tests:
- Encode: take `input`, encode it, compare it to `output`
- Decode: take `output`, decode it, compare it to `input`

include/veriblock/base58.h
AharonNavi: Fixed

include/veriblock/base58.h
AharonNavi: fixed

src/crypto/base58.cpp
AharonNavi: fixed

AharonNavi: fixed

test/base58_test.cpp
AharonNavi: fixed

include/veriblock/base58.h
AharonNavi: fixed

PR #5, opened 2020-01-10
Desserialization feature

AharonNavi:
implement base59


include/veriblock/base59.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include <string>
+#include <vector>
+
+namespace VeriBlock {
+
+    std::string EncodeBase59(const unsigned char *pbegin,
+        const unsigned char *pend);
+
+    // std::string EncodeBase58(const uint8_t *buf, size_t size);
```
Warchant: why commented?

src/base59.cpp
```diff
@@ -0,0 +1,150 @@
...
+  for (size_t i = startAt; i < number59.size(); i++) {
+    int digit59 = (int8_t)number59[i] & 0xFF;
+    size_t temp = remainder * g_kBase59 + digit59;
+    number59[i] = (uint8_t)(temp / g_kBase_256);
+    remainder = temp % g_kBase_256;
+  }
+  return (uint8_t)remainder;
+}
+
+std::string EncodeBase59(const void *buf, size_t nSize) {
```
Warchant:
```suggestion
std::string EncodeBase59(const uint8_t *buf, size_t nSize) {
```
anyway, we use this as byte array

src/base59.cpp
```diff
@@ -0,0 +1,150 @@
...
+  if (!nSize) {
+    return result;
+  }
+
+  // Make a copy of the input since we are going to modify it.
+  std::vector<uint8_t> input(nSize);
+  if (input.empty()) {
+    return result;
+  }
+  memcpy(input.data(), buf, nSize);
```
Warchant: to make a copy, do 
```
std::vector<uint8_t> input(buf, buf + size);
```

src/base59.cpp
```diff
@@ -0,0 +1,150 @@
...
+    temp[--j] = g_Base59Alphabet[mod];
+  };
+
+  // Strip extra '1' if any
+  while (j < temp.size() && temp[j] == g_Base59Alphabet.c_str()[0]) {
+    ++j;
+  }
+
+  // Add as many leading '1' as there were leading zeros.
+  for (--zeroCount; zeroCount != (size_t)-1; zeroCount--) {
```
Warchant:
```suggestion
  for (--zeroCount; zeroCount != std::numeric_limits<size_t>::max(); zeroCount--) {
```

src/base59.cpp
Warchant: UPD: actually, remove result from this place

src/base59.cpp
```diff
@@ -0,0 +1,150 @@
...
+      temp[--j] = g_Base59Alphabet[0];
+  }
+
+  result.resize(temp.size() - j);
```
Warchant: UPD: do this 
```
return std::vector<uint8_t>{temp.begin(), temp.end() - j};
```

src/base59.cpp
```diff
@@ -0,0 +1,150 @@
...
+  memcpy((void *)result.data(), &temp.data()[j], temp.size() - j);
+  return result;
+};
+
+std::string EncodeBase59(const unsigned char *pbegin, 
+    const unsigned char *pend) {
+    return EncodeBase59(pbegin, pend - pbegin);
```
Warchant: why `pend - pbegin`?

src/base59.cpp
```diff
@@ -0,0 +1,150 @@
...
+};
+
+std::vector<uint8_t> DecodeBase59(const std::string &input) {
+  std::vector<uint8_t> result;
+  if (input.empty()) {
+    return result;
+  }
+
+  memset(g_Indexes.data(), -1, g_Indexes.size() * sizeof(uint32_t));
```
Warchant: try to not use `memcpy`, `memset` and friends. use vector's constructors

also, why `g_Indexes` is not function-local?
when you declare it outside, you make this function thread-unsafe, because you do **writes** to global array

src/base59.cpp
```diff
@@ -0,0 +1,150 @@
...
+  for (size_t i = 0; i < g_Base59Alphabet.size(); i++) {
+    g_Indexes[g_Base59Alphabet[i]] = (uint32_t)i;
```
Warchant: `g_Indexes` can be generated and copied as array
https://github.com/VeriBlock/alt-integration-cpp/blob/master/src/base58.cpp#L38

src/base59.cpp
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  // Do no add extra leading zeroes, move j to first non null byte.
+  while (j < temp.size() && temp[j] == 0) {
+    ++j;
+  }
+  result.resize(temp.size() - j + zeroCount);
+  memcpy(result.data(), temp.data() + j - zeroCount,
+         temp.size() - j + zeroCount);
+  return result;
```
Warchant: this can be rewritten
```
return std::vector<uint8_t>{temp.data() + j - zeroCount, temp.data() - j + zeroCount};
```

include/veriblock/base59.hpp
AharonNavi: header has been copied from your base58.hpp, but 58 - postfix was changed to 59 - postfix. 
Delete this comments on base58 and base59?

src/base59.cpp
AharonNavi: Disagree. Result is std::string, is not std::vector

AharonNavi: I propose
return std::string{temp.begin(), temp.end() - j};

src/base59.cpp
AharonNavi: this function is wrapper under std::string EncodeBase59(const uint8_t *buf, size_t nSize)

src/base59.cpp
AharonNavi: Porting artefact. Agree.

src/base59.cpp
AharonNavi: fixed

src/base59.cpp
AharonNavi: fixed

src/base59.cpp
AharonNavi: fixed

src/base59.cpp
AharonNavi: fixed

src/base59.cpp
AharonNavi: fixed

src/base59.cpp
AharonNavi: fixed

include/veriblock/base59.hpp
AharonNavi: uncommented this function

src/base59.cpp
AharonNavi: fixed

src/base59.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  if (!nSize) {
+    return std::string();
+  }
+
+  // Make a copy of the input since we are going to modify it.
+  std::vector<uint8_t> input(buf, buf+nSize);
+  if (input.empty()) {
+    return std::string();
+  }
```
Warchant: this check is redundant, because you check that size is 0, and input can be 0 only if `nSize` is zero

src/base59.cpp
Warchant: use `size_t size = std::distance(pbegin, pend);` instead

src/base59.cpp
```diff
@@ -0,0 +1,146 @@
...
+  // Transform the String to a base59 byte sequence
+  for (size_t i = 0; i < input.size(); ++i) {
+    uint8_t c = input[i];
+
+    int digit59 = -1;
+    if (c >= 0 && c < 128) {
+      digit59 = g_Indexes[c];
+    }
+    if (digit59 < 0) {
+      throw new std::invalid_argument("Not a Base59 input");
```
Warchant:
```suggestion
      throw new std::invalid_argument("DecodeBase59(): Not a Base59 input");
```

src/base59.cpp
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    temp[--j] = mod;
+  }
+
+  // Do no add extra leading zeroes, move j to first non null byte.
+  while (j < temp.size() && temp[j] == 0) {
+    ++j;
+  }
+  return std::vector<uint8_t>{temp.data() + j - zeroCount, temp.data() + temp.size()};
```
Warchant:
```suggestion
  return std::vector<uint8_t>{temp.begin() + j - zeroCount, temp.end()};
```

test/base59_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+struct TestCase {
+    std::vector<uint8_t> binData;
+    std::string baseData;
+};
+
+class Base59Test : public testing::TestWithParam<TestCase> {};
+
+//  test data
+static std::vector<TestCase> g_Cases = {
+    {
```
Warchant: add some more test cases

https://github.com/VeriBlock/nodecore-js/tree/master/packages/core/testData

Here you can find two files. The one with `java` suffix, contains base59 encoded strings, that are sourced from `hex` file.

E.g. 

`base59(hex[0]) == java[0])`

Use `#include "util/literals.hpp" and `"012abcde"_unhex to get vector instead of hexstring.
E.g.
```
{"b4df3138d6b98a5a6350e653f4d6cb0c"_unhex, "GcJp9rfcTQfG104YwUGAk9"}
```

Also, use `_unhex` for current vectors - represent them as hexstring, and use `_unhex` - this makes reading much simpler

src/base59.cpp
AharonNavi: fixed

test/base59_test.cpp
AharonNavi: done

PR #15, opened 2020-01-16
Feature/entity vbktx btc 105

blackyblack:


include/veriblock/entities/vbktx.hpp
```diff
@@ -0,0 +1,109 @@
...
+    NetworkBytePair networkOrType = readNetworkByte(stream, TxType::VBK_TX);
+    Address sourceAddress = Address::fromVbkEncoding(stream);
+    Coin sourceAmount = Coin::fromVbkEncoding(stream);
+
+    uint8_t outputSize = stream.readBE<uint8_t>();
+    std::vector<Output> outputs;
+    outputs.reserve(outputSize);
+    for (size_t i = 0; i < outputSize; i++) {
+      outputs.emplace_back(Output::fromVbkEncoding(stream));
+    }
```
Warchant: can be replaced with `readArrayOf`

blackyblack: Don't think so. readArrayOf assumes length is stored with writeSingleByteLenValue.

include/veriblock/entities/vbkblock.hpp
```diff
@@ -3,35 +3,64 @@
...
+  int16_t version;
+  VBlakeBlockHash previousBlock;
+  VBlakePrevKeystoneHash previousKeystone;
+  VBlakePrevKeystoneHash secondPreviousKeystone;
+  VbkMerkleRootSha256Hash merkleRoot;
+  int32_t timestamp;
+  int32_t difficulty;
+  int32_t nonce;
+
+  VbkBlock(int32_t _height,
```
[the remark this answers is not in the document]
blackyblack: It was made this way before. I didn't like partially constructed objects but I agree that it looks better this way.

include/veriblock/shautil.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include "sha256.h"
+
+namespace VeriBlock {
+
+/**
+ * Calculates SHA256 of the input data
+ * @param data read data from this array
+ * @return vector with SHA256 hash of the input data
+ */
+Sha256Hash sha256get(Slice<uint8_t> data);
```
Warchant:
```suggestion
Sha256Hash sha256(Slice<uint8_t> data);
```

src/shautil.cpp
```diff
@@ -0,0 +1,17 @@
+#include "veriblock/shautil.hpp"
+
+namespace VeriBlock {
+
+Sha256Hash sha256get(Slice<uint8_t> data) {
+  uint8_t btctxHash[SHA256_HASH_SIZE]{};
```
Warchant:
```suggestion
  Sha256Hash btctxHash{};
```

Warchant: also add assert 
```
assert(data.size() <= std::numeric_limits<uint32_t>::max());
```

src/shautil.cpp
```diff
@@ -0,0 +1,17 @@
...
+  sha256(btctxHash, data.data(), (uint32_t)data.size());
```
Warchant: and then pass
```
btctxHash.data()
```

PR #20, opened 2020-01-21
Fix warnings

blackyblack:


src/secp256k1/secp256k1.c
```diff
@@ -216,36 +216,20 @@ void secp256k1_scratch_space_destroy(const secp256k1_context *ctx, secp256k1_scr
...
-        secp256k1_fe_set_b32(&x, pubkey->data);
-        secp256k1_fe_set_b32(&y, pubkey->data + 32);
-        secp256k1_ge_set_xy(ge, &x, &y);
-    }
+    /* When the secp256k1_ge_storage type is exactly 64 byte, use its
+     * representation inside secp256k1_pubkey, as conversion is very fast.
+     * Note that secp256k1_pubkey_save must use the same representation. */
+    secp256k1_ge_storage s;
+    memcpy(&s, &pubkey->data[0], sizeof(s));
+    secp256k1_ge_from_storage(ge, &s);
```
Warchant: I think this should be reverted, and warning should be ignored in this place.
sizeof(secp256k1_ge_storage) == 64 is always true in some OSes, but on other OSes will fail.

src/secp256k1/secp256k1.c
```diff
@@ -216,36 +216,20 @@ void secp256k1_scratch_space_destroy(const secp256k1_context *ctx, secp256k1_scr
...
-        secp256k1_ge_to_storage(&s, ge);
-        memcpy(&pubkey->data[0], &s, sizeof(s));
-    } else {
-        VERIFY_CHECK(!secp256k1_ge_is_infinity(ge));
-        secp256k1_fe_normalize_var(&ge->x);
-        secp256k1_fe_normalize_var(&ge->y);
-        secp256k1_fe_get_b32(pubkey->data, &ge->x);
-        secp256k1_fe_get_b32(pubkey->data + 32, &ge->y);
-    }
+    secp256k1_ge_storage s;
```
Warchant: same

src/secp256k1/secp256k1.c
```diff
@@ -288,26 +272,17 @@ int secp256k1_ec_pubkey_serialize(const secp256k1_context* ctx, unsigned char *o
...
-        /* When the secp256k1_scalar type is exactly 32 byte, use its
-         * representation inside secp256k1_ecdsa_signature, as conversion is very fast.
-         * Note that secp256k1_ecdsa_signature_save must use the same representation. */
-        memcpy(r, &sig->data[0], 32);
-        memcpy(s, &sig->data[32], 32);
-    } else {
-        secp256k1_scalar_set_b32(r, &sig->data[0], NULL);
-        secp256k1_scalar_set_b32(s, &sig->data[32], NULL);
-    }
+    /* When the secp256k1_scalar type is exactly 32 byte, use its
```
Warchant: same

PR #23, opened 2020-01-23
Feature/secp256k1 vbk support btc 190

blackyblack:


src/signutil.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+  throw std::invalid_argument("publicKeyFromVbk(): invalid public key");
+}
+
+std::vector<uint8_t> publicKeyToVbk(PublicKey key) {
+  return publicKeyUncompressedToAsn1(key);
+}
+
+PublicKey derivePublicKey(PrivateKey privateKey) {
+  checkRange(privateKey.size(), PRIVATE_KEY_SIZE, PRIVATE_KEY_SIZE);
```
Warchant: I feel like checkRange does not fit here.
better use exact length check 

```
if(privateKey.size() != PRIVATE_KEY_SIZE)
```

this semantically is more correct

src/signutil.cpp
```diff
@@ -0,0 +1,197 @@
...
+      &pubkey,
+      SECP256K1_EC_UNCOMPRESSED);
+  if (!res) {
+    throw std::invalid_argument(
+        "derivePublicKey(): public key serialize failed");
+  }
+  return output;
+}
+
+std::vector<uint8_t> veriBlockSign(Slice<uint8_t> message,
```
Warchant:
```suggestion
std::vector<uint8_t> veriBlockSign(Slice<const uint8_t> message,
```
here and in other places

the difference is same as between
```
uint8_t* 
and
const uint8_t*
```

src/signutil.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+  if (!ret) {
+    throw std::invalid_argument("sha256EcdsaSign(): cannot sign");
+  }
+
+  unsigned char sig[100]{};
+  size_t outputlen = 100;
+  secp256k1_ecdsa_signature_serialize_der(ctx(), sig, &outputlen, &signature);
+  std::vector<uint8_t> output(sig, sig + outputlen);
+  return output;
```
Warchant:
```suggestion
  return std::vector<uint8_t>{sig, sig+outputlen};
```

src/signutil.cpp
```diff
@@ -0,0 +1,197 @@
...
+  secp256k1_ec_pubkey_parse(ctx(), &pubkey, publicKey.data(), publicKey.size());
+
+  secp256k1_ecdsa_signature signature;
+  secp256k1_ecdsa_signature_parse_der(
+      ctx(), &signature, signatureEncoded.data(), signatureEncoded.size());
+
+  auto messageHash = sha256(message);
+
+  int ret = secp256k1_ecdsa_verify(ctx(), &signature, messageHash.data(), &pubkey);
+  return ret;
```
Warchant: return directly output of function

PR #24, opened 2020-01-25
Create third_party dir; add tinyfmt lib

Warchant:


include/veriblock/fmt.hpp
```diff
@@ -0,0 +1,24 @@
...
+ * Now it is tinyfmt, but can be any.
+ */
+
+namespace VeriBlock {
+
+/**
+ * A facade for a printf-like formatting.
+ */
+template <typename... Args>
+std::string format(const char* fmt, const Args&... args) {
```
blackyblack: Why do we need this? Ain't standard printf good enough?

Warchant: pay attention that this function does not **print** anything. It works as printf which writes into std::string 
```
std::string f = printf("%d", 50);
```

PR #27, opened 2020-01-27
Feature/check address btc 183

blackyblack:


include/veriblock/entities/address_entity.hpp
```diff
@@ -0,0 +1,35 @@
...
+#define ALT_INTEGRATION_ADDRESS_ENTITY_HPP
+
+#include <cstdint>
+#include <string>
+
+#include "veriblock/consts.hpp"
+
+namespace VeriBlock {
+
+struct AddressEntity {
```
Warchant: why renamed to AddressEntity?

and why removed existing serialization/deserialization functions?

src/address.cpp
```diff
@@ -0,0 +1,205 @@
...
+  }
+  return false;
+}
+
+static bool isBase59String(const std::string& input) {
+  try {
+    auto decoded = DecodeBase59(input);
+    (void)decoded;
+    return true;
+  } catch (std::invalid_argument e) {
```
Warchant:
```suggestion
  } catch (const std::invalid_argument& e) {
```

test/entities/address_test.cpp
```diff
@@ -12,32 +12,94 @@ static auto ADDRESS_BYTES =
...
+}
+
+TEST(Address, DerivedFromPublicKey) {
+  auto publicKey =
+      "3056301006072a8648ce3d020106052b8104000a03420004cb427e41a0114874080"
+      "a4b1e2ab7920e22cd2d188c87140defa447ee5fc44bb848e1c0db5ef206de2e7002"
+      "f6c86952be4823a4c08e65e4cdbeb904a8b95763aa"_unhex;
+  std::string addressString = "VFFDWUMLJwLRuNzH4NX8Rm32E59n6d";
+  AddressEntity address = addressFromString(addressString);
+  EXPECT_EQ(addressIsDerivedFromPublicKey(address, publicKey),
```
Warchant:
```suggestion
  EXPECT_TRUE(addressIsDerivedFromPublicKey(address, publicKey));
```

test/entities/address_test.cpp
```diff
@@ -12,32 +12,94 @@ static auto ADDRESS_BYTES =
...
+}
+
+TEST(Address, NotDerivedFromPublicKey) {
+  auto publicKey =
+      "3056301006072a8648ce3d020106052b8104000a03420004cb427e41a0114874080"
+      "a4b1e2ab7920e22cd2d188c87140defa447ee5fc44bb848e1c0db5ef206de2e7002"
+      "f6c86952be4823a4c08e65e4cdbeb904a8b95763aa"_unhex;
+  std::string addressString = "V23Cuyc34u5rdk9psJ86aFcwhB1md0";
+  AddressEntity address = addressFromString(addressString);
+  EXPECT_EQ(addressIsDerivedFromPublicKey(address, publicKey), false);
```
Warchant:
```suggestion
  EXPECT_FALSE(addressIsDerivedFromPublicKey(address, publicKey));
```

include/veriblock/entities/address_entity.hpp
blackyblack: I want to separate data structures and associated parse/serialize/validate functions. We either move everything to address.hpp or move everything out of there. I've chosen the latter.

src/address.cpp
blackyblack: Does not compile on my side.

src/address.cpp
```diff
@@ -0,0 +1,205 @@
...
+  auto data =
+      std::string{"V"} + keyHashEncoded.substr(0, MULTISIG_ADDRESS_DATA_END);
+  auto checksum = calculateChecksum(data, false);
+  return AddressEntity(AddressType::STANDARD, data + checksum);
+}
+
+bool addressIsDerivedFromPublicKey(const AddressEntity& address,
+                                   Slice<const uint8_t> publicKey) {
+  auto expectedAddress = addressFromPublicKey(publicKey);
+  if (!(address == expectedAddress)) {
```
Warchant:
```suggestion
  if (address != expectedAddress) {
```

src/address.cpp
```diff
@@ -0,0 +1,205 @@
...
+    return true;
+  } catch (std::invalid_argument e) {
+    // do not throw - return status instead
+  }
+  return false;
+}
+
+static std::string calculateChecksum(const std::string& data, bool multisig) {
+  std::vector<uint8_t> dataBytes = toBytes(data);
+  auto hash = sha256(dataBytes);
```
Warchant: no need to copy string to vector and then calculate sha256 from vector. you can calculate sha256 directly from string

PR #31, opened 2020-01-29
Add block repository interface

Warchant:


include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,165 @@
...
+ * @invariant any instance is in valid state - database is opened when valid
+ * instance of BlockRepository created. To close database, delete the instance.
+ */
+template <typename Block>
+struct BlockRepository {
+  //! stored block type
+  using stored_block_t = Block;
+  //! block has type
+  using hash_t = typename Block::hash_t;
+  //! block heiehgt type
```
blackyblack: height

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,165 @@
...
+   * batch.
+   */
+  struct WriteBatch {
+    virtual ~WriteBatch() = default;
+
+    /**
+     * Write a single block. If block with such hash exists, db will overwrite
+     * it.
+     * @param block to be written in a batch
+     * @return true if block already existed in db and we overwrote it. False
```
blackyblack: true if we overwrite existing block

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,165 @@
...
+   * @param height block height
+   * @return true if at least one block removed, false otherwise.
+   */
+  virtual int removeByHeight(height_t height) = 0;
+
+  /**
+   * Create new WriteBatch, to perform BULK modify operations.
+   * @return a pointer to new WriteBatch instance.
+   */
+  virtual std::unique_ptr<WriteBatch> getBatch() = 0;
```
blackyblack: How about newBatch() name?

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * Efficiently commit given batch on-disk.
+   * @param batch
+   */
+  virtual void commit(WriteBatch& batch) = 0;
```
blackyblack: Does it throw any exception? Execution result?

include/veriblock/storage/block_repository.hpp
```diff
@@ -16,6 +16,9 @@ namespace VeriBlock {
  * Multiple blocks may be stored at the given height.
  *
  * @tparam Block Stored Block type
+ *
+ * @invariant any instance is invalid state - database is opened when valid
```
eu321: in a valid

include/veriblock/storage/block_repository.hpp
Warchant: It is implementation defined.

include/veriblock/storage/block_repository.hpp
Warchant: fixed

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+   * batch.
+   */
+  struct WriteBatch {
+    virtual ~WriteBatch() = default;
+
+    /**
+     * Write a single block. If block with such hash exists, db will overwrite
+     * it.
+     * @param block to be written in a batch
+     * @return true if we overwrite existing block, false otherwise.
```
eu321: can sevral writebatches coexist in an uncommitted state? what would put return in such a situation?

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+     */
+    virtual bool put(const stored_block_t& block) = 0;
+
+    /**
+     * Remove a single block from storage identified by its hash.
+     * @param hash block hash
+     * @return true if removed, false if no such element found.
+     */
+    virtual bool removeByHash(const hash_t& hash) = 0;
```
eu321: an alternative would be to have several overloaded remove() functions

include/veriblock/storage/block_repository.hpp
Warchant: you can create as many batches as you want, until you commit them - nothing happens

include/veriblock/storage/block_repository.hpp
Warchant: yes, I just added move verbosity :)

Warchant: yes, I just added move verbosity :)

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+  virtual bool getByHash(const hash_t& hash, stored_block_t* out) = 0;
+
+  /**
+   * Load potentially many blocks from disk in memory by their height.
+   * @param height[in] block height
+   * @param out[out] if non-null, blocks will be written here. If null passed,
+   * out argument is ignored.
+   * @return true if block found at given height, false otherwise.
+   */
+  virtual bool getByHeight(const height_t& height,
```
eu321: int& doesn't make much sense

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+                           std::vector<stored_block_t>* out) = 0;
+
+  /**
+   * Load many blocks from disk in memory by a list of hashes.
+   * @param hashes[in] a list of hashes to load.
+   * @param out[out] if non-null, blocks will be written here. If null passed,
+   * out argument is ignored.
+   * @return true if at least one block has been found, false otherwise.
+   */
+  virtual bool getManyByHash(Slice<const hash_t> hashes,
```
eu321: it will be quite hard to figure out what blocks were found/not found. returning a bool would probably not help much.
the api spec doesn't say if the vector is appended to or overwritten

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+                             std::vector<stored_block_t>* out) = 0;
+
+  /**
+   * Write a single block. If block with such hash exists, db will overwrite
+   * it.
+   * @param block to be written in a batch
+   * @return true if block already existed in db and we overwrote it. False
+   * otherwise.
+   */
+  virtual void put(const stored_block_t& block) = 0;
```
eu321: he comment above states this should be bool not void

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+   * @return true if removed, false if no such element found.
+   */
+  virtual bool removeByHash(const hash_t& hash) = 0;
+
+  /**
+   * Remove potentially many blocks at given height.
+   * @param height block height
+   * @return true if at least one block removed, false otherwise.
+   */
+  virtual int removeByHeight(height_t height) = 0;
```
eu321: true if at least one block removed-> probably this was intended to return the number of blocks removed

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+   * Create new WriteBatch, to perform BULK modify operations.
+   * @return a pointer to new WriteBatch instance.
+   */
+  virtual std::unique_ptr<WriteBatch> newBatch() = 0;
+
+  /**
+   * Efficiently commit given batch on-disk.
+   * @param batch
+   */
+  virtual void commit(WriteBatch& batch) = 0;
```
eu321: probably should take unique_ptr<WriteBatch>. so deleting the newBatch() output causes a transaction rollback and Commit() ensures there's proper clean-up and no undefined state(eg commit a batch, add more ops to the batch, commit again).

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+     * @return true if removed, false if no such element found.
+     */
+    virtual bool removeByHash(const hash_t& hash) = 0;
+
+    /**
+     * Remove potentially many blocks at given height.
+     * @param height block height
+     * @return true if at least one block removed, false otherwise.
+     */
+    virtual bool removeByHeight(height_t height) = 0;
```
eu321: probably should return int = the removed block count

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * Load potentially many blocks from disk in memory by their height.
+   * @param height[in] block height
+   * @param out[out] if non-null, blocks will be written here. If null passed,
+   * out argument is ignored.
+   * @return true if block found at given height, false otherwise.
+   */
+  virtual bool getByHeight(const height_t& height,
+                           std::vector<stored_block_t>* out) = 0;
```
eu321: the api doc doesn't specify whether the vector is appended to or cleared by the function

eu321: if the vector is appended to, it might make more sense to return the block count, and not just a bool

include/veriblock/storage/block_repository.hpp
eu321: if the vector is appended to, it might make more sense to return the block count

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+   * otherwise.
+   */
+  virtual void put(const stored_block_t& block) = 0;
+
+  /**
+   * Remove a single block from storage identified by its hash.
+   * @param hash block hash
+   * @return true if removed, false if no such element found.
+   */
+  virtual bool removeByHash(const hash_t& hash) = 0;
```
eu321: the api fails to cover a potentially important use case, deleting a block and getting a copy of it. think of a pop stack operation

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+  };
+
+  /**
+   * Load a block from disk in memory, by its hash.
+   * @param hash[in] block hash
+   * @param out[out] if non-null, block data will be written here. If null
+   * passed, out argument is ignored.
+   * @return true if block found, false otherwise.
+   */
+  virtual bool getByHash(const hash_t& hash, stored_block_t* out) = 0;
```
eu321: this API looks soooo C-like. it's definitely asking for a pattern like Block get(Hash) and letting the compiler optimize the object creation/copy/move code.
if it is left like this, I'd at least add  a default =nullptr value for out or better yet, or provide 2 overloads, one that lacks out(could name it exists() while we're at it) and another where out is a reference

include/veriblock/storage/block_repository.hpp
eu321: ok not it, but probably something compatible with vector sizes and unsigned like size_t

include/veriblock/storage/block_repository.hpp
eu321: but if we attempt to act upon the put output when we have several batches that touch an overlapping set of blocks, we might get a nasty result

include/veriblock/storage/block_repository.hpp
eu321: re: lack of optional -- returning unique_ptr is an option but it wouldn't be as efficient as returning the block itself. Block get(Hash) that throws an exception if the block is not found might also make some sense. yes, my suggestions obviously cause the number of functions to increase, but that's what you'd need to get the compiler to do more static checking of your code

eu321: although, we could always play the custom allocator/resource pool game with unique_ptr, so pretty interface != slow interface. but it does take work

eu321: another option would be to return shared_ptr<const Block> and reuse the cached block instances

include/veriblock/storage/block_repository.hpp
```diff
@@ -0,0 +1,164 @@
...
+   * @param height block height
+   * @return number of blocks removed
+   */
+  virtual size_t removeByHeight(height_t height) = 0;
+
+  /**
+   * Create new WriteBatch, to perform BULK modify operations.
+   * @return a pointer to new WriteBatch instance.
+   */
+  virtual std::unique_ptr<WriteBatch> newBatch() = 0;
```
eu321: newBatch could take an optional capacity hint just like vector does

include/veriblock/storage/block_repository.hpp
Warchant: you can execute `commit` only sequentially. It would be very hard to mess up.

include/veriblock/storage/block_repository.hpp
Warchant: there will be no undefined state in a batch. If you want that - you can do that.

include/veriblock/storage/block_repository.hpp
Warchant: We always can add more functions when we need them

include/veriblock/storage/block_repository.hpp
Warchant: This API is very similar to what RocksDB or LevelDB offer. I think we will stick to this API. It is low level and flexible enough to be usable. Also it is easy to implement and use.

include/veriblock/storage/block_repository.hpp
Warchant: there's no reason to do that. 
if implementation need that, we can add new method.

PR #32, opened 2020-01-29
Constant Slice reverse

blackyblack:


include/veriblock/slice.hpp
```diff
@@ -64,10 +64,9 @@ struct Slice {
     return const_reverse_iterator{cbegin()};
   }
 
-  Slice<ElementType> reverse() const {
-    Slice<ElementType> ret = *this;
-    std::reverse(const_cast<value_type *>(ret.begin()),
-                 const_cast<value_type *>(ret.end()));
+  std::vector<value_type> reverse() const {
+    std::vector<value_type> ret(begin(), end());
```
Warchant: is there any other std::reverse overloading that can help to omit unnecessary copy?
something like
```
 // only one array pass instead of 2 
return std::vector<value_type>{rbegin(), rend()}; 
```

PR #42, opened 2020-02-11
Feature/blockchain generic test btc 214

blackyblack:


include/veriblock/blob.hpp
```diff
@@ -40,8 +40,11 @@ struct Blob {
     data_.fill(0);
     assign(str);
   }
+  Blob(Blob<N>& other) {
```
Mr-Leshiy:
```suggestion
template<typename M>
  Blob(const Blob<M>& other) {
```

include/veriblock/blob.hpp
```diff
@@ -40,8 +40,11 @@ struct Blob {
...
+    data_.fill(0);
```
Warchant: no need to do this before `assign`

include/veriblock/entities/vbkblock.hpp
```diff
@@ -55,6 +59,30 @@ struct VbkBlock {
...
 
+  /*
+   * TODO
+   */
+  uint32_t getDifficulty() const;
+
+  /*
+   * TODO
+   */
+  uint32_t getBlockTime() const;
```
Warchant: here it is just a getter for `timestamp`

include/veriblock/entities/vbkblock.hpp
```diff
@@ -55,6 +59,30 @@ struct VbkBlock {
    */
   void toVbkEncoding(WriteStream& stream) const;
 
+  /*
+   * TODO
+   */
+  uint32_t getDifficulty() const;
```
Warchant: it is just a getter for `difficulty`

test/blockchain/blockchain_test.cpp
```diff
@@ -0,0 +1,242 @@
...
+    EXPECT_CALL(*cursor, seekToFirst()).Times(1);
+    // database is empty, so first isValid will return false
+    EXPECT_CALL(*cursor, isValid()).WillOnce(Return(false));
+
+    // @when
+    auto result = blockchain->bootstrap(height, genesis, state);
+    EXPECT_TRUE(result);
+    EXPECT_TRUE(state.IsValid());
+  };
+};
```
Warchant: these two can be combined into single templated BlockchainInstance, and then parametrized with according types:
```
struct BtcTestCase {
  using params_t = BtcChainParams;
  using hash_t = decltype(BtcBlock::previousBlock);
  using block_t = BtcBlock;
  using height_t = int;
};

struct VbkTestCase {
  using params_t = VbkChainParams;
  using hash_t = decltype(VbkBlock::previousBlock);
  using block_t = VbkBlock;
  using height_t = int;
};
```

include/veriblock/blockchain/blocktree.hpp
```diff
@@ -68,7 +68,9 @@ struct BlockTree {
       return state.addStackFunction("acceptBlockHeader()");
     }
 
-    auto it = block_index_.find(block.previousBlock);
+    // make a copy of hash to let it extend to hash_t size
+    hash_t previousBlockHash(block.previousBlock);
+    auto it = block_index_.find(previousBlockHash);
```
Warchant: why is it needed?

Warchant:
Overall looks good :)

include/veriblock/blockchain/blocktree.hpp
blackyblack: Auto conversion does not work here. Probably could be solved with proper Blob constructor but could not find it.

include/veriblock/entities/vbkblock.hpp
blackyblack: Yes but we should keep it documented.

include/veriblock/blob.hpp
blackyblack: Will it properly assign smaller size blob?

Warchant: Hm. I don't think so.

PR #44, opened 2020-02-11
[1] Add inmem block repo

Warchant:
- remove "{remove,get}ByHeight" methods from BlockRepo
- add BlockRepositoryInmemory + tests
- TODO: update rocksdb to meet new interface (rocksdb is disabled)


include/veriblock/storage/block_repository.hpp
```diff
@@ -60,26 +62,25 @@ struct WriteBatch {
    * @param block to be written in a batch
    * @return true if we overwrite existing block, false otherwise.
```
blackyblack: Does not return bool anymore

include/veriblock/storage/block_repository.hpp
```diff
@@ -60,26 +62,25 @@ struct WriteBatch {
...
    */
-  virtual bool put(const stored_block_t& block) = 0;
+  virtual void put(const stored_block_t& block) = 0;
 
   /**
    * Remove a single block from storage identified by its hash.
    * @param hash block hash
    * @return true if removed, false if no such element found.
```
blackyblack: Same

include/veriblock/storage/block_repository.hpp
```diff
@@ -60,26 +62,25 @@ struct WriteBatch {
...
-  virtual size_t removeByHeight(height_t height) = 0;
+  virtual void clear() = 0;
 
   /**
-   * Clear batch from any modifying operations.
+   * Efficiently commit given batch on-disk. Clears batch from changes.
+   * @param repo
    */
-  virtual void clear() = 0;
+  virtual void commit(BlockRepository<Block>& repo) = 0;
```
blackyblack: Shouldn't it be reverse way? A BlockRepository with commit(WriteBatch) method.

Warchant: hard to implement it this way... :-(

Warchant: because repo has access only to parent class "WriteBatch", and without implementation details it is hard

PR #52, opened 2020-02-13
Conditionally find rocksdb package

blackyblack:


src/CMakeLists.txt
```diff
@@ -26,6 +26,10 @@ add_library(vbkutils
...
 
 if(WITH_ROCKSDB)
-    target_link_libraries(vbkutils rocksdb)
-    target_include_directories(vbkutils PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../deps/rocksdb/include>)
+    if(FIND_ROCKSDB)
+        target_link_libraries(vbkutils RocksDB::rocksdb)
+    else()
+        target_link_libraries(vbkutils rocksdb)
+        target_include_directories(vbkutils PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../deps/rocksdb/include>)
+    endif()
```
Warchant: Create alias for target, and use alias instead of condition.
https://cmake.org/cmake/help/v3.3/command/add_library.html#id5

Also, add target_include_directories to `rocksdb`, so that all libs that link `rocksdb` (from submodule) get this include dir passed.

PR #54, opened 2020-02-14
add getNextWorkRequired realisation for the VbkBlock

Mr-Leshiy:


include/veriblock/blockchain/block_index.hpp
```diff
@@ -48,6 +48,23 @@ struct BlockIndex {
     return pbegin[(pend - pbegin) / 2];
   }
 
+  const BlockIndex* getAncestorBlocksBehind(height_t steps) const {
+    if (steps < 0 || steps > this->height) {
+      return nullptr;
+    }
+
+    if (steps == 0) {
```
blackyblack: Checks for steps are not necessary. Code will work the same way with an assumption that steps <= 0 return current block.

test/blockchain/vbk_blockchain_util_test.cpp
```diff
@@ -0,0 +1,191 @@
...
+#include "veriblock/blockchain/block_index.hpp"
+#include "veriblock/blockchain/vbk_blockchain_util.hpp"
+#include "veriblock/blockchain/vbk_chain_params.hpp"
+
+using namespace VeriBlock;
+
+static std::vector<BlockIndex<VbkBlock>> getChain(int32_t deltaTime,
+                                                  int32_t deltaTime_change,
+                                                  const uint32_t& difficulty,
+                                                  const uint32_t& chainlength) {
```
blackyblack: chainlength and difficulty should be passed by value.

test/blockchain/vbk_blockchain_util_test.cpp
```diff
@@ -0,0 +1,191 @@
...
+  BlockIndex<VbkBlock> blockIndex;
+  blockIndex.header.height = 1;
+  blockIndex.header.timestamp = 10000;
+  blockIndex.header.difficulty = difficulty;
+  blockIndex.height = blockIndex.header.height;
+  blockIndex.pprev = nullptr;
+
+  std::vector<BlockIndex<VbkBlock>> chain(chainlength);
+  chain[0] = blockIndex;
```
blackyblack: If chainlength is zero an exception will be thrown here. Put an assert to make sure chainlength is not zero.

test/blockchain/vbk_blockchain_util_test.cpp
```diff
@@ -0,0 +1,191 @@
...
+  int32_t deltaTime = chainparams->getTargetBlockTime();
+
+  BlockIndex<VbkBlock> blockIndex;
+  blockIndex.header.height = 1;
+  blockIndex.header.timestamp = 10000;
+  blockIndex.header.difficulty = ArithUint256::fromHex("09184E72A000").toBits();
+  blockIndex.height = blockIndex.header.height;
+  blockIndex.pprev = nullptr;
+
+  std::vector<BlockIndex<VbkBlock>> chain(chainlength);
```
blackyblack: I think you could use getChain() here.

Mr-Leshiy: Here is used another algorithm of the chain building that is hard to unify to all purposes of the previous tests

include/veriblock/blockchain/block_index.hpp
Mr-Leshiy: fixed

test/blockchain/vbk_blockchain_util_test.cpp
Mr-Leshiy: fixed

test/blockchain/vbk_blockchain_util_test.cpp
Mr-Leshiy: fixed

PR #55, opened 2020-02-16
Generic storage test

blackyblack:


include/veriblock/storage/block_repository_rocks.hpp
```diff
@@ -5,13 +5,107 @@
...
+                          hash.size());
+    rocksdb::Status s = _batch.Delete(_hashBlockHandle.get(), blockHash);
+    if (!s.ok() && !s.IsNotFound()) {
+      throw db::DbError(s.ToString());
+    }
+  }
+
+  void clear() override { _batch.Clear(); }
+
+  void commit(BlockRepository<Block>& /* ignore */) override {
```
Warchant: why ignore?

test/storage/storage_test.cpp
```diff
@@ -0,0 +1,266 @@
...
+#include "veriblock/storage/block_repository_inmem.hpp"
+#include "veriblock/storage/block_repository_rocks_setup.hpp"
+#include "veriblock/uint.hpp"
+
+using namespace VeriBlock;
+
+enum class TEST_RUNS { INMEM = 0, ROCKS_BTC, ROCKS_VBK };
+
+struct BlockBasic {
+  using hash_t = uint256;
```
Warchant: why did you change simple-to-use `int` to hard-to-use `hash256`?

test/storage/storage_test.cpp
```diff
@@ -0,0 +1,266 @@
...
+        break;
+    }
+    repo->clear();
+  }
+
+ protected:
+  BlockRepositoryRocksInstance<BlockBasic, BlockBasic> database;
+  std::shared_ptr<BlockRepositoryInmem<BlockBasic>> inmemRepo;
+  std::shared_ptr<BlockRepositoryRocks<BlockBasic>> rocksBtcRepo;
+  std::shared_ptr<BlockRepositoryRocks<BlockBasic>> rocksVbkRepo;
```
Warchant: why do you have all 3 here?

include/veriblock/storage/block_repository_rocks.hpp
blackyblack: We do not have any way to access DB instance from the BlockRepository interface. Only way is to pass DB ref during WriteBatch creation. Therefore we do not need BlockRepository to commit batch to the DB.

test/storage/storage_test.cpp
blackyblack: Because RocksDB assumes keys as byte arrays. Integers cannot be accessed the same way - they do not have neither data() nor size().

test/storage/storage_test.cpp
blackyblack: I tried to put them outside the Test class but for some reason RocksDB is not initialized properly this way.

test/storage/storage_test.cpp
blackyblack: I updated the code to use getHash() from the block interface and convert int to hash_t for BasicBlock. Code is cleaner now.

include/veriblock/storage/block_repository_rocks.hpp
Warchant: maybe remove argument?

include/veriblock/storage/block_repository_rocks.hpp
```diff
@@ -171,31 +195,23 @@ class BlockRepositoryRocks : public BlockRepository<Block> {
...
+  void clear() override { 
+    // call BlockRepositoryRocksManager.clear() instead
+    return;
+  }
 
-      result.insert(Blob<32>(blobPartStr));
-    }
-    return result;
+  std::unique_ptr<WriteBatch<stored_block_t>> newBatch() override {
+    return std::make_unique<WriteBatchRocks<stored_block_t>>(_db,
```
Warchant: make unique is C++14 feature https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

Use `std::unique_ptr<T>(new T)`

include/veriblock/storage/block_repository_rocks.hpp
blackyblack: We will have to change the interface as well. So BlockRepositoryInmem will be changed accordingly. Do you want to do it?

include/veriblock/storage/block_repository_rocks.hpp
blackyblack: +

include/veriblock/storage/block_repository_rocks.hpp
Warchant: If it is easy to do, yeah

PR #59, opened 2020-02-20
Add comparePopScore

Warchant:


src/blockchain/pop/fork_resolution.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <veriblock/consts.hpp>
+#include <veriblock/keystone_util.hpp>
+
+namespace VeriBlock {
+
+namespace {
+
+bool publicationViolatesFinality(int pubToCheck, int base) {
+  int diff = pubToCheck - base;
+  return diff - BITCOIN_FINALITY_DELAY > 0;
```
blackyblack: diff might be negative here. Maybe add < 0 check?

Warchant: this is okay.
negative - BITCOIN_FINALITY_DELAY will never be > 0

PR #61, opened 2020-02-21
add endorsement_repository_inmem realisation with tetst

Mr-Leshiy:


src/storage/endorsements_repository_inmem.cpp
```diff
@@ -0,0 +1,58 @@
+#include "veriblock/storage/endorsements_repository_inmem.hpp"
+
+using namespace VeriBlock;
+
+void EndorsementsRepositoryInmem::put(
+    const VbkPopTx& popTx, const VbkBlock::hash_t& containingBlockHash) {
+  VbkBlock::hash_t endorsedBlockHash = popTx.publishedBlock.getHash();
+  std::string popTxHash = popTx.getHash().asString();
```
Warchant: i think you can replace std::string with `decltype(popTx.getHash())`

src/storage/endorsements_repository_inmem.cpp
```diff
@@ -0,0 +1,58 @@
...
+      break;
+    }
+  }
+  if (endrs.empty()) {
+    endorsed_map[endorsedBlockHash] = {{
+        popTx.getHash().asString(),
+        popTx.publishedBlock.getHash(),
+        {containingBlockHash},
+        popTx.blockOfProof.getHash(),
+        popTx.address.toString(),
```
Warchant: we store address, but do we use it?

src/storage/endorsements_repository_inmem.cpp
```diff
@@ -0,0 +1,58 @@
...
+  std::vector<Endorsement>& endrs = endorsed_map[endorsedBlockHash];
+  auto endorsement = endrs.begin();
+  for (; endorsement != endrs.end(); ++endorsement) {
+    if (endorsement->endorsementKey == popTxHash) {
+      endorsement->containingVbkHashes.push_back(containingBlockHash);
+      break;
+    }
+  }
+  if (endrs.empty()) {
+    endorsed_map[endorsedBlockHash] = {{
```
Warchant: no need to do that. when you execute `endorsed_map[endorsedBlockHash]`, map creates new element and returns it.
so technically you can just push to `endrs`

src/storage/endorsements_repository_inmem.cpp
Mr-Leshiy: as I remember, no.

Warchant: remove it, then

Mr-Leshiy: No I have mistaken, we will able to use it in the pop reward algorithm

src/storage/endorsements_repository_inmem.cpp
```diff
@@ -0,0 +1,50 @@
...
+void EndorsementsRepositoryInmem::put(
+    const VbkPopTx& popTx, const VbkBlock::hash_t& containingBlockHash) {
+  VbkBlock::hash_t endorsedBlockHash = popTx.publishedBlock.getHash();
+  std::string popTxHash = popTx.getHash().asString();
+
+  std::vector<Endorsement>& endrs = endorsed_map[endorsedBlockHash];
+  auto endorsement = endrs.begin();
+  for (; endorsement != endrs.end(); ++endorsement) {
+    if (endorsement->endorsementKey == popTxHash) {
+      endorsement->containingVbkHashes.push_back(containingBlockHash);
```
Warchant: somewhere here you also need to check that `endorsedVbkHash` also belongs to chain (call func on it)

src/storage/endorsements_repository_inmem.cpp
```diff
@@ -0,0 +1,50 @@
...
+    const VbkBlock::hash_t& endorsedBlockHash,
+    chainContainsBlockFunc func) const {
+  std::vector<Endorsement> result;
+
+  if (endorsed_map.empty()) {
+    return result;
+  }
+
+  for (const auto& endorsement : endorsed_map.at(endorsedBlockHash)) {
+    for (const auto& containingHashes : endorsement.containingVbkHashes) {
```
Warchant: somewhere here you also need to check that `endorsedVbkHash` also belongs to chain (call func on it)

PR #63, opened 2020-02-21
Set time

Warchant:


src/blockchain/btc_blockchain_util.cpp
```diff
@@ -99,10 +99,49 @@ uint32_t getNextWorkRequired(const BlockIndex<BtcBlock>& prevBlock,
...
       prevBlock, pindexFirst->getBlockTime(), params);
 }
 
+template <>
+int64_t getMedianTimePast(const BlockIndex<BtcBlock>& prev) {
+  static constexpr int medianTimeSpan = 11;
+
+  int64_t pmedian[medianTimeSpan];
+  std::fill(pmedian, pmedian + medianTimeSpan, 0);
+  auto* pbegin = &pmedian[medianTimeSpan];
```
blackyblack: If prev BlockIndex is null pbegin will point to pmedian.end(). Therefore any access to pbegin will throw an exception.

src/blockchain/vbk_blockchain_util.cpp
```diff
@@ -110,4 +112,55 @@ uint32_t getNextWorkRequired(const BlockIndex<VbkBlock>& prevBlock,
...
+  const BlockIndex<VbkBlock>* pindex = &prev;
+  for (i = 0; i < HISTORY_FOR_TIMESTAMP_AVERAGE; i++, pindex = pindex->pprev) {
+    if (pindex == nullptr) {
+      break;
+    }
+    pmedian.push_back(pindex->getBlockTime());
+  }
+
+  std::sort(pmedian.begin(), pmedian.end());
+  size_t index = i % 2 == 0 ? (i / 2) - 1 : (i / 2);
```
blackyblack: For i = 0 the result will be a very big integer.

src/blockchain/vbk_blockchain_util.cpp
```diff
@@ -110,4 +112,55 @@ uint32_t getNextWorkRequired(const BlockIndex<VbkBlock>& prevBlock,
...
+  return pmedian.at(index);
```
blackyblack: Unfilled pmedian will fail here for any index.

src/blockchain/btc_blockchain_util.cpp
Warchant: how prev can be null? it is passed as reference in this function

src/blockchain/vbk_blockchain_util.cpp
Warchant: i can not be 0, because `prev` is passed by reference, and i will be at least 1

src/blockchain/vbk_blockchain_util.cpp
Warchant: this will never ever happen

PR #65, opened 2020-02-26
implement "disconnectTip" in BlockTree

Mr-Leshiy:


include/veriblock/blockchain/blocktree.hpp
```diff
@@ -262,10 +343,14 @@ struct BlockTree {
                                   index_t& indexNew) {
     if (currentBest.tip() == nullptr ||
         currentBest.tip()->chainWork < indexNew.chainWork) {
+      auto prevTip = currentBest.tip();
       currentBest.setTip(&indexNew);
+      addForkCandidate(prevTip, &indexNew);
```
blackyblack: oldCandidate is a second parameter. Why do you put new tip as an oldCandidate?

src/blockchain/pop/vbk_block_tree.cpp
```diff
@@ -139,12 +139,15 @@ void VbkBlockTree::determineBestChain(Chain<block_t>& currentBest,
 
   if (result > 0) {
     // other chain won!
-    return currentBest.setTip(&indexNew);
+    auto prevTip = currentBest.tip();
+    currentBest.setTip(&indexNew);
+    return addForkCandidate(prevTip, &indexNew);
```
blackyblack: Also prevTip as a new candidate.

include/veriblock/blockchain/blocktree.hpp
```diff
@@ -109,10 +109,54 @@ struct BlockTree {
...
+    if (blockIndex == nullptr) {
+      return state.Invalid("invalidateBlockByHash()",
+                           "bad-block-hash",
+                           "can not find block by this hash");
+    }
+
+    for (auto it = fork_candidates_.begin(); it != fork_candidates_.end();) {
+      index_t* forkIndex = *it;
+      index_t* prevIndex = nullptr;
+      if (forkIndex->getAncestor(blockIndex->height) == blockIndex) {
```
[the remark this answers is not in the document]
blackyblack: I think there is another issue as well. We might erase a branch having a common block with another branch. This way another branch will become orphaned.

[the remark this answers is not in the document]
blackyblack: I would recommend to mark branches containing block that's being deleted first. And physically erase branches on the next step.

include/veriblock/blockchain/blocktree.hpp
```diff
@@ -109,10 +109,32 @@ struct BlockTree {
     return acceptBlock(block, state, true);
   }
 
+  void invalidateBlockByHash(const hash_t& blockHash) {
+    index_t* blockIndex = getBlockIndex(blockHash);
```
Warchant: here you should check that block index is not nullptr.
`getBlockIndex` returns nullptr when it can't find a block with this hash

include/veriblock/blockchain/blocktree.hpp
```diff
@@ -148,15 +169,74 @@ struct BlockTree {
       current->height = current->pprev->height + 1;
       current->chainWork = current->pprev->chainWork + getBlockProof(block);
     } else {
-      current->height = 0;
```
Warchant: why did you remove that?

include/veriblock/blockchain/blocktree.hpp
```diff
@@ -148,15 +169,74 @@ struct BlockTree {
...
     return current;
   }
 
+  void invalidateBlockFromChain(Chain<Block>& chain, index_t** block) {
+    if (*block == nullptr) {
+      return;
+    }
+
+    if (!chain.contains(*block) &&
+        chain.tip()->getAncestor((*block)->height) != *block) {
```
Warchant: why do you need this check here?
isn't `contains` enough?

include/veriblock/blockchain/chain.hpp
```diff
@@ -57,9 +57,11 @@ struct Chain {
     return (*this)[index->height + 1];
   }
 
-  height_t size() const { return (height_t)chain.size() + startHeight_; }
+  height_t height() const { return (height_t)chain.size() + startHeight_ - 1; }
```
Warchant: when chain is empty, and startHeight is 0, height will point to -1, which is wrong. Better use `size() -1` in places where you need `height` of the chain

include/veriblock/blockchain/miner.hpp
```diff
@@ -15,7 +14,8 @@ struct Miner {
   using merkle_t = decltype(Block::merkleRoot);
   using index_t = BlockIndex<Block>;
 
-  Miner(std::shared_ptr<ChainParams> params) : params_(std::move(params)) {}
+  Miner(std::shared_ptr<ChainParams> params, uint32_t startTime)
```
Warchant:
```suggestion
  Miner(std::shared_ptr<ChainParams> params, uint32_t startTime = currentTimestamp4())
```

include/veriblock/blockchain/blocktree.hpp
```diff
@@ -109,10 +109,32 @@ struct BlockTree {
     return acceptBlock(block, state, true);
   }
 
+  void invalidateBlockByHash(const hash_t& blockHash) {
+    index_t* blockIndex = getBlockIndex(blockHash);
+
+    for (auto chain_it = fork_chains_.begin();
+         chain_it != fork_chains_.end();) {
+      invalidateBlockFromChain(*chain_it, &blockIndex);
```
blackyblack: In case common block of two branches is deleted one of the branches becomes orphaned.

include/veriblock/blockchain/blocktree.hpp
```diff
@@ -148,15 +169,74 @@ struct BlockTree {
...
+      if (chain_it->tip() == newCandidate) {
+        isAdded = true;
+      }
+
+      if (chain_it->tip() == oldCandidate) {
+        chain_it = fork_chains_.erase(chain_it);
+        continue;
+      }
+      ++chain_it;
+    }
```
blackyblack: It seems there are two not connected actions: deleting oldCandidate and adding newCandidate. I think they should reside in two different functions.

include/veriblock/blockchain/blocktree.hpp
Mr-Leshiy: no it is not enough, because fork chains dont have all blocks, and it is on that case that we want do remove the old block from the main chain, and as a result we have to remove the parts of the forks.

include/veriblock/blockchain/chain.hpp
Mr-Leshiy: I am not sure, because the meaning of the size is not the similar with the realisation, it is not return the size of the chain.

include/veriblock/blockchain/blocktree.hpp
Warchant: instead of getAncestor use `activeChain.contains(*block)`

include/veriblock/blockchain/blocktree.hpp
```diff
@@ -147,10 +177,93 @@ struct BlockTree {
...
+      return;
+    }
+
+    // find block from the main chain
+    index_t* workBlock = newCandidate;
+    if (forkChainStart == fork_chains_.end()) {
+      for (; !activeChain_.contains(workBlock->pprev);
+           workBlock = workBlock->pprev)
+        ;
+    }
```
Warchant: use `auto* forkBlock = chain.findFork(tip)`

include/veriblock/blockchain/blocktree.hpp
```diff
@@ -107,10 +107,58 @@ struct BlockTree {
...
+    // and if block F is being removed, and chain 'F - G - H - I' will be
+    // processed firstly, we will remove block 'G' which is needed for the
+    // another chain, so that chain must be processed firstly
+
+    std::sort(
+        fork_chains_.begin(),
+        fork_chains_.end(),
+        [](const Chain<Block>& chain1, const Chain<Block>& chain2) -> bool {
+          return chain1.getStartHeight() > chain2.getStartHeight();
+        });
```
Warchant: use std::map to store forks. it will maintain sorted order (by height)

PR #67, opened 2020-02-27
Remove load from Block tree, add contextuallyCheckBlock

Warchant:


include/veriblock/blockchain/blocktree.hpp
```diff
@@ -157,45 +153,11 @@ struct BlockTree {
...
-      }
-    }
-
-    return true;
-  }
-
   bool acceptBlock(const block_t& block,
                    ValidationState& state,
-                   bool validateBlock) {
+                   bool checkDifficulty) {
```
blackyblack: This is not checkDifficulty anymore. It is checkDifficulty for BTC and checkByPrevious for VBK. Advised to rename to something more generic or split to two parameters.

include/veriblock/blockchain/pop/vbk_block_tree.hpp
```diff
@@ -19,9 +19,8 @@ struct VbkBlockTree : public BlockTree<VbkBlock, VbkChainParams> {
 
   VbkBlockTree(BtcTree& btc,
                std::shared_ptr<EndorsementsRepository> erepo,
```
blackyblack: Shouldn't erepo be moved out as well?

Warchant: we use as read-only repo for fork resolution

include/veriblock/blockchain/blocktree.hpp
Warchant: Fixed

include/veriblock/blockchain/pop/vbk_block_tree.hpp
Warchant: another possible way to solve it is to pass a function instead of repo, as we need only single function...

PR #85, opened 2020-03-24
change StateManager and StateChange

Mr-Leshiy:


include/veriblock/state_manager.hpp
```diff
@@ -114,14 +57,14 @@ class StateManager {
   RepositoryManager database;
 
  public:
-  StateManager(const std::string& name) : database(name) { database.open(); }
+  StateManager(const std::string& name, std::string path = "")
+      : database(path + "\\" + name) {
```
blackyblack: Why this change? We could add separators outside, by the StateManager users.

src/popmanager.cpp
```diff
@@ -84,49 +82,41 @@ bool PopManager::addVTB(const VTB& vtb,
...
-        stateChange.removeBtcBlock(b.getHash());
-      });
+  std::for_each(btcctx.rbegin(), btcctx.rend(), [this](const BtcBlock& b) {
+    btc_->invalidateBlockByHash(b.getHash());
+  });
 }
 
 bool PopManager::addAltProof(const AltProof& payloads,
-                             StateChange& stateChange,
+                             StateChange&,
```
blackyblack: How about deleting this?

Mr-Leshiy: It is already deleted in the feature/alt-tree branch

include/veriblock/state_manager.hpp
Mr-Leshiy: fixed

PR #108, opened 2020-04-06
[WIP] AltTree test

Mr-Leshiy:


src/mock_miner.cpp
```diff
@@ -36,58 +36,93 @@ bool mineBlocks(const uint32_t& n,
...
+  }
+
+  // since we inserted in reverse order, we need to reverse context blocks
+  std::reverse(atv.context.begin(), atv.context.end());
+
+  if (!vbktree.acceptBlock(containingBlock, {}, state)) {
+    throw std::domain_error(state.GetPath() + "\n" + state.GetDebugMessage());
+  }
+
+  altPayload.alt.atv = atv;
```
Warchant, line 103: you only generate one ATV?
You need to generate 1 per AltTx

src/mock_miner.cpp
```diff
@@ -36,58 +36,93 @@ bool mineBlocks(const uint32_t& n,
...
+    const VbkBlock::hash_t& lastKnownVbkBlockHash,
+    const BtcBlock::hash_t& lastKnownBtcBlockHash,
+    uint32_t number_of_vtbs,
+    ValidationState& state) {
+  AltPayloads altPayload;
+
+  // build merkle tree
+  auto hashes = hashAll<VbkTx>({transaction});
+  const int32_t treeIndex = 0;  // this is POP tx
+  VbkMerkleTree mtree(hashes, treeIndex);
```
Warchant, line 72: This functionality should be inside `mineAltBlock`. 
When you want to generate X alt transactions in single alt block, you need to create one block per tree, and that merkle tree has to include all txes

PR #123, opened 2020-04-16
overload acceptBlock

Mr-Leshiy:


Warchant:
Please fix CI

include/veriblock/entities/context.hpp
```diff
@@ -11,19 +11,57 @@ namespace altintegration {
...
+  VbkBlock getContainingBlock() const;
+
+  /**
+   * Return true if contains endorsement data
+   * @return true if contains endorsement data
+   */
+  bool containsEndorsements() const;
+
+  /**
+   * Return BtcEndorsement of the PartialVbk
```
blackyblack, line 40: of the PartialVTB

include/veriblock/entities/context.hpp
```diff
@@ -11,19 +11,57 @@ namespace altintegration {
...
+
+  /**
+   * Return BtcEndorsement id of the endorsement
+   * @return endorsement.id BtcEndorsement::id_t
+   */
+  typename BtcEndorsement::id_t getEndorsementId() const;
+
+  // generate without VBK context blocks, only BtcEndorsement VbkBlock
+  // containing and btc blocks
+  static PartialVTB generateFromVtb(const VTB& vtb);
```
blackyblack, line 53: If PartialVTB does not contain VBK context blocks why do you have context_vbk in the struct?

src/entities/context.cpp
```diff
@@ -0,0 +1,61 @@
...
+  std::transform(vtbs.begin(),
+                 vtbs.end(),
+                 std::back_inserter(p_vtbs),
+                 [&](const VTB& vtb) -> PartialVTB {
+                   PartialVTB p_vtb = PartialVTB::generateFromVtb(vtb);
+                   for (const auto& b : vtb.context) {
+                     p_vtb.context_vbk.push_back(std::make_shared<VbkBlock>(b));
+                   }
+                   return p_vtb;
+                 });
```
blackyblack, line 35: It does not look safe. When generating PartialVTBs from the vector, you populate PartialVTB.context_vbk, but when you generate PartialVTB from the single VTB, context_vbk stays uninitialised.

include/veriblock/entities/context.hpp
Mr-Leshiy: This context blocks fills independently in the other place, and that context can be different with the context blocks from the VTB, so in this place I don not copy them.

src/entities/context.cpp
```diff
@@ -0,0 +1,61 @@
...
+  return endorsement.id;
+}
+
+bool operator==(const PartialVTB& a, const PartialVTB& b) {
+  bool is_equal = a.btc.size() == b.btc.size();
+  for (size_t i = 0; i < a.btc.size() && is_equal; ++i) {
+    if (*a.btc[i] != *b.btc[i]) {
+      is_equal = false;
+    }
+  }
```
Warchant, line 56: I think it should be enough to use regular ==

src/entities/context.cpp
```diff
@@ -0,0 +1,61 @@
...
+    p_vtb.btc.push_back(std::make_shared<BtcBlock>(b));
+  }
+  p_vtb.btc.push_back(std::make_shared<BtcBlock>(vtb.transaction.blockOfProof));
+
+  p_vtb.containing = std::make_shared<VbkBlock>(vtb.containingBlock);
+
+  return p_vtb;
+}
+
+std::vector<PartialVTB> PartialVTB::generateFromVtb(
```
Warchant, line 21:
```suggestion
std::vector<PartialVTB> PartialVTB::fromVTB(
```

src/entities/vtb.cpp
```diff
@@ -43,8 +43,12 @@ std::vector<uint8_t> VTB::toVbkEncoding() const {
 }
 
 VTB::id_t VTB::getId() const {
-  auto rawBytes = toVbkEncoding();
-  return sha256(rawBytes);
+  WriteStream stream;
+  transaction.toVbkEncoding(stream);
+  merklePath.toVbkEncoding(stream);
```
Warchant, line 48: VTB should not have merkle path as part of ID

PR #127, opened 2020-04-17
BAD IDEA: add a signature normalization workaround that introduces a signature malleability vulnerability

eu321:


Warchant:
Do not merge, I am adding test for this patch

PR #128, opened 2020-04-18
Add scenario 0

Warchant:


include/veriblock/blockchain/pop/fork_resolution.hpp
```diff
@@ -373,7 +373,15 @@ struct PopAwareForkResolutionComparator {
   const protected_index_t* getIndex() const { return index_; }
 
   void removePayloads(protected_index_t& index,
-                      const std::vector<protected_payloads_t>& payloads) {
+                      const std::vector<protected_payloads_t>& payloads,
+                      int64_t lastApplied = -1) {
```
blackyblack, line 377: I recommend some better name for lastApplied. Maybe endWith or lastIndex.

include/veriblock/blockchain/pop/fork_resolution.hpp
```diff
@@ -462,7 +473,7 @@ struct PopAwareForkResolutionComparator {
 
           return true;
         },
-        [&] { removePayloads(index, payloads); });
+        [&] { removePayloads(index, payloads, lastApplied); });
```
blackyblack, line 476: More logical solution is to maintain a local vector of added payloads and pass it to the removePayloads instead of using additional lastApplied index.

src/blockchain/pop/vbk_block_tree.cpp
```diff
@@ -150,7 +150,8 @@ void VbkBlockTree::removePayloads(index_t* index,
                                   const std::vector<payloads_t>& payloads) {
   assert(index);
   cmp_.removePayloads(*index, payloads);
-  if (index->containingContext.back().empty()) {
+  if (!index->containingContext.empty() &&
+      index->containingContext.back().empty()) {
```
blackyblack, line 154: Why this additional check? Is it possible that back() is empty? Why we don't allow to pop this empty element?

test/e2e/scenario_0/scenario_0.cpp
```diff
@@ -0,0 +1,116 @@
...
+  while (std::getline(s, line)) {
+    if (ParseHex(line).empty()) {
+      break;
+    }
+    ret.push_back(transform(line));
+  }
+  return ret;
+}
+
+struct Scenario0 : public ::testing::Test {
```
blackyblack, line 41: Why is it Scenario0? Is it some basic, very important scenario to make it before scenario1?

test/e2e/scenario_0/scenario_0.cpp
```diff
@@ -0,0 +1,116 @@
...
+    config->btc.startHeight = btcStart;
+    config->btc.blocks =
+        parse<BtcBlock>(generated::btcbootstraps, fromHex<BtcBlock>);
+
+    config->vbk.params = std::make_shared<VbkChainParamsTest>();
+    config->vbk.startHeight = vbkStart;
+    config->vbk.blocks =
+        parse<VbkBlock>(generated::vbkbootstraps, fromHex<VbkBlock>);
+
+    auto t = Altintegration::create(*config);
```
blackyblack, line 70: Altintegration::create - where did it come from?

src/blockchain/pop/vbk_block_tree.cpp
Warchant: it is possible that deque will not have items, and `context.back()` fails because of that

test/e2e/scenario_0/scenario_0.cpp
Warchant: Yeah, it contains real data from vBTC :)

test/e2e/scenario_0/scenario_0.cpp
Warchant: there's a factory class `Altintegration`, it accepts a config and does bootstraps of all trees in correct order

PR #134, opened 2020-04-22
macos fix; add macos to ci

overcookedpanda:
One change was required for this to compile on macos, and I also added macos to the ci, however, clang-tidy is not available with homebrew, so if you want to run that test on macos, you'll want to either install all of llvm with brew and create the symbolic links to clang-tidy or download the binaries and link the same.


Warchant:
You don't need to add macos to matrix build. Just add one build in Release (`-DCMAKE_BUILD_TYPE=Release), this should be enough.

PR #136, opened 2020-04-23
Implement Command pattern for undo

Warchant:


Warchant:
Continuing work in feature/cmd

PR #138, opened 2020-04-24
add license to CPP/H/HPP all content

xagau:


xagau:
license to all content (CPP/H/HPP)  merged.

PR #139, opened 2020-04-27
Add test for pop payouts

Warchant:


Warchant:
@blackyblack Approved

PR #140, opened 2020-04-28
Add validity flags to BlockIndex

Warchant:
Do not remove blocks from BlockTree when doing InvalidateBlock

### DIFF
--- a/include/veriblock/blockchain/block_index.hpp
+++ b/include/veriblock/blockchain/block_index.hpp
@@ -137,7 +137,8 @@ struct BlockIndex {
 
   std::string toPrettyString() const {
     return "BlockIndex{height=" + std::to_string(height) +
-           ", hash=" + getHash().toHex().substr(0, 8) +
+           ", hash=" + HexStr(getHash()) +
+           ", prev=" + (pprev ? HexStr(pprev->getHash()) : "<empty>") +
            ", endorsedBy=" + std::to_string(endorsedBy.size()) +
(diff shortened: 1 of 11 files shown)


include/veriblock/blockchain/block_index.hpp
```diff
@@ -19,6 +19,25 @@
...
+  //! block is statelessly valid, but we marked it as failed
+  BLOCK_FAILED_BLOCK = 4,
+  //! block failed POP validation
+  BLOCK_FAILED_POP = 8,
+  //! block is state{lessly,fully} valid, but some of previous blocks is invalid
+  BLOCK_FAILED_CHILD = 16,
+  //! all invalidity flags
+  BLOCK_FAILED_MASK = BLOCK_FAILED_CHILD | BLOCK_FAILED_POP | BLOCK_FAILED_BLOCK
+};
+
```
blackyblack, line 40: It seems it does not matter if it is BLOCK_FAILED_CHILD or BLOCK_FAILED_BLOCK or any other reason for a failure. Therefore in the long run these options will be messed up and used arbitrarily. I recommend to use a single FAILED flag.

Warchant: In current code we distinguish blocks that we marked as invalid manually (BLOCK_FAILED_BLOCK), or blocks that for descend from such block (BLOCK_FAILED_CHILD).
This will be used in future code, in command processing...

PR #163, opened 2020-05-06
Fix/mempool tests

Mr-Leshiy:


src/blockchain/alt_block_tree.cpp
```diff
@@ -219,6 +219,8 @@ void payloadsToCommands<AltTree>(AltTree& tree,
     commands.push_back(std::move(cmd));
   }
 
+  addBlock(tree.vbk(), p.altPopTx.atv.containingBlock, commands);
```
Warchant, line 222: move this block undef `if` behind

Mr-Leshiy: fixed

PR #168, opened 2020-05-07
Feature/fix and remove rocksdb

overcookedpanda:


overcookedpanda:
I removed the git sub module for gtest, and then used CMake to clone and download the dependency and it solves the issue you ran into here https://github.com/VeriBlock/alt-integration-cpp/commit/eb8d5128a8df6ee7eb96cec042f1c78ed71dc722 with the coverage breaking.

PR #176, opened 2020-05-11
pyPopMiner

overcookedpanda:
This PR should fully integration pyPopMiner and the tests should run with each build, however, due to some Boost warnings, we may have to revise to run this particular build without -Werror


pypopminer/CMakeLists.txt
```diff
@@ -12,19 +12,21 @@ if(NOT ${Boost_FOUND})
 endif()
 find_package(PythonInterp 3)
 find_package(PythonLibs 3)
-add_custom_target(pypopminer COMMENT "Building pypopminer")
+add_custom_target(src COMMENT "Building pyPopMiner")
```
Warchant, line 15: Why is target named src?

overcookedpanda: IDE got carried away.. fixes incoming :)

pypopminer/src/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+include_directories(${PROJECT_SOURCE_DIR}/src)
```
Warchant, line 2: Why do you add this?

overcookedpanda: `set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})` was existing in your original commits, somewhere through the debugging process, I added `include_directories(${PROJECT_SOURCE_DIR}/src)` but it seems both can be removed now.  Will commit fix now.

PR #179, opened 2020-05-13
Feature/add logs

Warchant:


include/veriblock/blockchain/blocktree.hpp
```diff
@@ -188,20 +191,20 @@ struct BlockTree : public BaseBlockTree<Block> {
...
-      return state.Invalid("check-block");
+      return state.Invalid(block_t::name() + "check-block");
     }
 
     // we must know previous block
     auto* prev = base::getBlockIndex(block->previousBlock);
     if (prev == nullptr) {
       return state.Invalid(
-          "bad-prev-block",
+          block_t::name() + "bad-prev-block",
```
blackyblack, line 201: I think it should be like block_t::name() + "-bad-prev-block" otherwise it would become unreadable.
The same applies for other snippets below.

src/blockchain/alt_block_tree.cpp
```diff
@@ -102,17 +107,23 @@ bool AltTree::validatePayloads(const AltBlock::hash_t& block_hash,
...
   }
 
   if (!setState(*index, state)) {
+    VBK_LOG_DEBUG("{} Statefully invalid payloads: {}",
+                  block_t::name(),
+                  state.toString());
     removePayloads(*index, {p});
     index->unsetFlag(BlockStatus::BLOCK_FAILED_POP);
-    return state.Invalid("addPayloadsTemporarily");
+    return state.Invalid(block_t::name() + "addPayloadsTemporarily");
```
blackyblack, line 126: Use the same format as above, eg add-payloads-temporarily

PR #188, opened 2020-05-20
Feature/fix rollback

eu321:


include/veriblock/blockchain/pop/pop_state_machine.hpp
```diff
@@ -96,7 +122,7 @@ struct PopStateMachine {
 
     for (auto* index : chain) {
       if (!index->isValid() || !applyBlock(*index, state)) {
-        unapply(*index, from);
+        unapply(*index->pprev, from);
```
Warchant, line 125: you can not do that, because index is not guaranteed to be in the middle of chain, it can be very first block

include/veriblock/blockchain/commands/addblock.hpp
```diff
@@ -35,12 +35,12 @@ struct AddBlock : public Command {
   void UnExecute() override {
     auto hash = block_->getHash();
     auto* index = tree_->getBlockIndex(hash);
-    if (!index) {
-      return;
-    }
+    assert(index != nullptr
```
Warchant, line 38: UnExecute is not guaranteed to be executed after command applied. It will be executed on AddBlock if it failed. And if so, this block will not exist in blockchain, so nothing to remove here. `index` will be nullptr here.

include/veriblock/blockchain/pop/pop_state_machine.hpp
Warchant: also chain may have size=1

eu321: the chain starts at from.height + 1, thus index->pprev of the first block in the chain == from; unapply(from, from) is a no-op.

include/veriblock/blockchain/commands/addblock.hpp
eu321: Reference counting blocks only makes sense if Execute and UnExecute calls match.Otherwise it becomes possible for example to execute a command group that adds blocks 1,2,3(eg the BTC context of a VTB), and then unexecute a different command group that adds blocks 1 and 2(the context shipped by another VTB)

include/veriblock/blockchain/base_block_tree.hpp
```diff
@@ -92,6 +92,13 @@ struct BaseBlockTree {
     }
   }
 
+  void removeTip(index_t& toRemove, bool shouldDetermineBestChain = true) {
+    if (!toRemove.pnext.empty()) {
```
Warchant, line 96:
```suggestion
    if (!toRemove.isValidTip()) {
```

Block may have next blocks and they may be invalid, so technically current block is tip

eu321: not from the point of view of rollbacks. those invalid blocks must still be deleted by other UnExecute calls

eu321: that is, one block per UnExecute

PR #215, opened 2020-06-03
fix the endorsement time adjustment

eu321:
This breaks lots of tests. We can get the tests to pass by shifting timestamps of  genesis blocks: altchain << vbk << btc . But this would merely conceal the fact that tests are still broken.


Warchant:
Fixed in #265

PR #218, opened 2020-06-04
Add ToJSON

Warchant:
By default we ship header only JSON library - picojson.
If users want, they can write their own adapter to embed any other JSON library.

### DIFF
--- a/include/veriblock/entities/vbk_merkle_path.hpp
+++ b/include/veriblock/entities/vbk_merkle_path.hpp
@@ -52,7 +52,7 @@ JsonObject ToJSON(const VbkMerklePath& mp) {
   json::putIntKV(obj, "treeIndex", mp.treeIndex);
   json::putIntKV(obj, "index", mp.index);
   json::putStringKV(obj, "subject", mp.subject.toHex());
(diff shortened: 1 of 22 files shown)


include/veriblock/entities/vbk_merkle_path.hpp
```diff
@@ -46,6 +46,16 @@ struct VbkMerklePath {
   uint128 calculateMerkleRoot() const;
 };
 
+template <typename JsonValue>
+JsonValue ToJSON(const VbkMerklePath& mp) {
+  JsonValue obj = json::makeEmptyObject<JsonValue>();
+  json::putIntKV(obj, "treeIndex", mp.treeIndex);
+  json::putIntKV(obj, "index", mp.index);
+  json::putStringKV(obj, "subject", mp.subject.toHex());
```
blackyblack, line 54: It was HexStr() in the MerklePath serialization above.

Warchant: These are same functions, so here it does not matter. It may matter with hashes, as AltBlock has vector<uint8_t> as hash, and others have Blob<N> as hash

PR #221, opened 2020-06-08
fix pop rewards issue

Mr-Leshiy:


Warchant:
This is not a solution.
Logged issue in JIRA https://veriblock.atlassian.net/browse/BTC-542

PR #223, opened 2020-06-08
Restore containingEndorsements map

blackyblack:


include/veriblock/blockchain/block_index.hpp
```diff
@@ -167,6 +169,11 @@ struct BlockIndex {
     return nullptr;
   }
 
+  bool isExistingPid(const pid_t& pid) const {
```
Warchant, line 172: more efficient would be to create a set of ids from this vector, and then try to insert into that set, and if item is inserted, then there's no duplicate.

```
std::set<pid_t> pids;
if(pids.insert(pid).second) {
  // can be inserted, unique element
} else {
  // duplicate
}
```
```

File: include/veriblock/blockchain/pop/pop_state_machine.hpp [hunk 2]
```diff
@@ -26,19 +26,34 @@ struct PopStateMachine {
...
 
   bool applyBlock(index_t& index, ValidationState& state) {
     std::vector<CommandPtr> executed;
-    for (auto& cg : index.commands) {
+    for (const auto& pid : index.payloadIds) {
+      payloads_t payloads;
+      bool ret = storage_.loadPayloads(pid, payloads);
+      if (!ret) {
+        //TODO: error message
+        continue;
```
Warchant, line 40: throw here. something like 
```
struct StateCorruptionException: public std::domain_error {
  using base = std::domain_error;
  using base::base; // to inherit constructors
};
```

Warchant: even better if loadPayloads will not return true/false on each payloads, but rather work in batches, and if loading is not possible, then it would throw StateCorruptionException:
```
applyBlock(...) {
  std::vector<CommandGroup> groups = storage_.loadCommands(index.payloadIds);
  // ... normal flow
  
}
```

include/veriblock/blockchain/pop/pop_state_machine.hpp
```diff
@@ -26,19 +26,34 @@ struct PopStateMachine {
...
+                    pid.toHex(),
                     index.toShortPrettyString());
-      for (auto& cmd : cg) {
+
+      for (auto& cmd : commands) {
         if (!cmd->Execute(state)) {
           // invalidate command group
-          cg.valid = false;
+          payloads.valid = false;
+          ret = storage_.savePayloads(payloads);
```
Warchant, line 52: you do not have to update storage here. we only update storage periodically from `save` function

include/veriblock/entities/vtb.hpp
```diff
@@ -19,11 +19,13 @@ namespace altintegration {
...
 struct VTB {
   using id_t = uint256;
+  using containing_block_t = VbkBlock;
 
   VbkPopTx transaction{};
   VbkMerklePath merklePath{};
-  VbkBlock containingBlock{};
+  containing_block_t containingBlock{};
   std::vector<VbkBlock> context{};
+  bool valid{};
```
Warchant, line 28: pls add comment

include/veriblock/blockchain/commands/addvtb.hpp
```diff
@@ -25,11 +26,28 @@ struct AddVTB : public Command {
...
+    // duplicates not allowed
+    if (std::find(index->payloadIds.begin(),
+                  index->payloadIds.end(),
+                  vtb_.getId()) != index->payloadIds.end()) {
+      return state.Invalid(
+          block_t::name() + "-duplicate-payloads",
+          fmt::sprintf("Containing block=%s already contains payload %s.",
+                       index->toPrettyString(),
+                       vtb_.getId().toHex()));
+    }
```
Warchant, line 45: addPayloads should handle duplicates by itself, why is this needed?

include/veriblock/blockchain/pop/fork_resolution.hpp
```diff
@@ -206,7 +207,7 @@ std::vector<ProtoKeystoneContext<ProtectingBlockT>> getProtoKeystoneContext(
       // chain must contain relevantEndorsedBlock
       VBK_ASSERT(index != nullptr);
 
-      for (const auto* e : index->endorsedBy) {
+      for (const auto e : index->endorsedBy) {
```
Warchant, line 210: ?

include/veriblock/blockchain/block_index.hpp
blackyblack: Why do you think it is more efficient? I see that constructing set from vector takes O(NlogN) and vector lookup is O(N).

include/veriblock/blockchain/pop/pop_state_machine.hpp
blackyblack: How do we add valid flag then?

include/veriblock/blockchain/commands/addvtb.hpp
blackyblack: Not sure it handles them properly. In scenario5 they were added without any error message but failed when tried to erase endorsedBy multiple times.

include/veriblock/blockchain/pop/pop_state_machine.hpp
blackyblack: Commands are constructed with tree reference. Should we pass payloadsToCommands lambda to storage_.loadCommands()?

include/veriblock/blockchain/pop/vbk_block_tree.hpp
```diff
@@ -83,6 +92,7 @@ struct VbkBlockTree : public BlockTree<VbkBlock, VbkChainParams> {
                           bool isBootstrap = false) override;
 
   PopForkComparator cmp_;
+  PayloadsStorage& storagePayloads_;
```
Warchant, line 95: we should probably change this to `std::shared_ptr<PayloadsStorage> storagePayloads_`

Warchant: same for ALT

PR #232, opened 2020-06-12
setState: fix atomicity: rollback on an attempt to switch to an invalid chain

eu321:
…


Warchant:
@eu321 good job! Pls make same PR to master

PR #240, opened 2020-06-16
tests: setState atomicity

eu321:


test/e2e/setstate_atomicity.cpp
```diff
@@ -0,0 +1,78 @@
...
+ * Mine chain B: endorse 1 block, add an invalid VTB to this block
+ * Activate chain A
+ * Attempt to activate chain B
+ * Check that the chain B activation attempt failed and left the state unchanged
+ */
+
+struct SetStateAtomicity : public ::testing::Test, public PopTestFixture {};
+
+TEST_F(SetStateAtomicity, setStateAtomicity) {
+  AltTree testAlttree = AltTree(altparam, vbkparam, btcparam);
```
Warchant, line 23: you could use `this->alttree` which is defined and bootstrapped in PopTestFixture

PR #252, opened 2020-06-24
fix state corruption in AddVTB on hitting the history overwrite limit and on attempting to add payloads to an invalid block; make addPayloads atomic; don't create invalid blocks in the mock miner

eu321:
…


eu321:
forgot to change
  /**
   * @invariant NOT atomic. When addPayloads is called and fails -
   * removePayloads have to be called on same payloads.
   */

PR #259, opened 2020-06-26
AddVTB: don't copy VTBs; misc cleanup

eu321:


include/veriblock/blockchain/commands/addvtb.hpp
```diff
@@ -18,32 +17,31 @@ struct AddVTB : public Command {
   using block_t = VbkBlock;
   ~AddVTB() override = default;
 
-  AddVTB(AltTree& tree, const VTB& vtb) : tree_(&tree), vtb_(vtb) {
-    id_ = vtb.getId().getLow64();
-  }
+  template <class VTB_t>
```
Warchant, line 20: Why template?

eu321: standard forwarding pattern

PR #281, opened 2020-07-06
add a test that forks are not validated completely

eu321:


Warchant:
Do not merge yet. We need to wait for CI on vbtc https://github.com/VeriBlock/vbk-ri-btc/pull/229/checks

PR #282, opened 2020-07-07
check that all the applied blocks always form a tree and no block is double-applied or double-unapplied; optimize the un(application) for BTC

eu321:
…

### DIFF
--- a/include/veriblock/blockchain/pop/pop_state_machine.hpp
+++ b/include/veriblock/blockchain/pop/pop_state_machine.hpp
@@ -37,69 +37,103 @@ struct PopStateMachine {
 
   // atomic: applies either all or none of the block's commands
   bool applyBlock(index_t& index, ValidationState& state) {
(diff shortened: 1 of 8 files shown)


include/veriblock/blockchain/pop/pop_state_machine.hpp
```diff
@@ -86,11 +97,32 @@ struct PopStateMachine {
...
       ed_.revalidateSubtree(index, BLOCK_FAILED_POP, /*do fr=*/false);
     }
+    index.setFlag(BLOCK_APPLIED);
     return true;
   }
 
   // atomic: applies either all of the block's commands or fails on an assert
-  void unapplyBlock(const index_t& index) {
+  void unapplyBlock(index_t& index) {
+    if (block_t::emptyPayloads) {
```
Warchant, line 106: unapply block is never called on BTC blocks, so this looks redundant

PR #285, opened 2020-07-07
Feature/command group cache 2 btc 576

blackyblack:


include/veriblock/command_group_cache.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+  virtual void clear() {
+    _cache.clear();
+    _keys.clear();
+    _priority.clear();
+  }
+
+ protected:
+  size_t _maxsize;
+  std::list<id_t> _priority;
```
Warchant, line 64: use std::vector

include/veriblock/command_group_cache.hpp
```diff
@@ -0,0 +1,93 @@
...
+  std::unordered_map<id_t, std::list<id_t>::iterator> _keys;
```
Warchant, line 65:
```suggestion
  std::unordered_map<id_t, std::vector<id_t>::iterator> _keys;
```

Warchant: std::list is very slow, instead would be better to use something like circular buffer built on top of std::vector

include/veriblock/blockchain/vbk_block_addon.hpp
```diff
@@ -27,6 +27,12 @@ struct VbkBlockAddon :
 
   bool payloadsIdsEmpty() const { return vtbids.empty(); }
 
+  template <typename pop_t, typename pop_id_t>
+  std::vector<pop_id_t>& getPayloadIds();
```
Mr-Leshiy, line 31: Why is it so important to specify this method with the pop_t type ?  Why does not enough only pop_id_t ?

include/veriblock/storage/payloads_storage.hpp
```diff
@@ -65,44 +77,120 @@ class PayloadsStorage {
...
+    return _cgValidity.find(cid) != _cgValidity.end();
+  }
+
  protected:
   std::shared_ptr<PayloadsRepository<ATV>> _repoAtv;
   std::shared_ptr<PayloadsRepository<VTB>> _repoVtb;
   std::shared_ptr<PayloadsRepository<VbkBlock>> _repoBlocks;
+  CommandGroupCache _cacheAlt;
+  CommandGroupCache _cacheVbk;
+  std::unordered_map<CommandGroup::id_t, bool> _cgValidity;
```
Warchant, line 178: Shouldn't it be `std::vector, bool`?
where vector is `containing block hash + command group id` ?

blackyblack: Yes. But it doesn't matter since CommandGroup::id_t is a vector itself.

PR #290, opened 2020-07-09
CI: Add initial support for MinGW build

overcookedpanda:


.github/workflows/ci.yml
```diff
@@ -152,6 +152,39 @@ jobs:
 #      - name: test
 #        run: cmake --build build --target RUN_TESTS
 
+  x86_64-w64-mingw32:
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        flags: ["-DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=x86_64-w64-mingw32-gcc -DCMAKE_CXX_COMPILER=x86_64-w64-mingw32-g++ -DCMAKE_SYSTEM_NAME=Windows -DWITH_ROCKSDB=ON -DTESTING=OFF"]
```
Warchant, line 160: if it is only single build, you can remove matrix build

overcookedpanda: It is a single build for now, but as we expand, we may need to include more options, so I left the structure as is, like in the other builds.

PR #291, opened 2020-07-09
Feature/remove should determine best chain set tip

eu321:


include/veriblock/blockchain/alt_block_tree.hpp
```diff
@@ -68,6 +68,13 @@ struct AltTree : public BaseBlockTree<AltBlock> {
   //! may return false, if bootstrap block is invalid
   bool bootstrap(ValidationState& state);
 
+  /**
+   * Check if the blockchain is bootstrapped
+   *
+   * @return true if the blockchain is bootstrapped, false otherwise
+   */
+  bool isBootstrapped();
```
Warchant: Should be moved to base_block_tree

src/blockchain/alt_block_tree.cpp
```diff
@@ -460,37 +463,47 @@ bool AltTree::addPayloads(AltTree::index_t& index,
...
-                      : "<empty>"));
-    activeChain_.setTip(&to);
-    tryAddTip(&to);
+  if (success) {
+    bool success = overrideTip(to, state);
+    VBK_ASSERT(success);
   } else {
+    // if setState failed, then 'to' must be invalid
     VBK_ASSERT(!to.isValid());
   }
```
Warchant, line 489: no need in this condition, because `success` is always true

eu321: setState is definitely not always going to succeed

Warchant: look closely - this is setTipContinueOnInvalid, it always succeeds, and there are two asserts that indicate this
https://github.com/VeriBlock/alt-integration-cpp/pull/291/files#diff-2cf3b8cf328ee87b5b03c1b7d0abf949R481
https://github.com/VeriBlock/alt-integration-cpp/pull/291/files#diff-2cf3b8cf328ee87b5b03c1b7d0abf949R485

eu321: ah right. github doesn't show enough context lines :(

PR #292, opened 2020-07-09
Add SaveTree, LoadTree

Warchant:


include/veriblock/blockchain/base_block_tree.hpp
```diff
@@ -239,9 +299,11 @@ struct BaseBlockTree {
...
 
   bool operator==(const BaseBlockTree& o) const {
-    TreeFieldsComparator cmp{};
-    return cmp(blocks_, o.blocks_) && cmp(tips_, o.tips_) &&
-           (activeChain_ == o.activeChain_);
+    CollectionOfPtrComparator cmp{};
+    bool a = cmp(blocks_, o.blocks_);
+    bool b = cmp(tips_, o.tips_);
+    bool c = (activeChain_ == o.activeChain_);
+    return a && b && c;
```
blackyblack, line 306: We might avoid unnecessary comparisons using the previous code.

include/veriblock/blockchain/block_index.hpp
```diff
@@ -189,13 +189,14 @@ struct BlockIndex : public Block::addon_t {
     return fromRaw(stream);
   }
 
-  friend bool operator==(const BlockIndex& a, const BlockIndex& b) {
-    return *a.header == *b.header;
+  bool operator==(const BlockIndex& o) const {
+    bool a = *header == *o.header;
+    bool b = status == o.status;
+    bool c = addon_t::operator==(o);
+    return a && b && c;
```
blackyblack, line 196: Same here.

include/veriblock/blockchain/btc_block_addon.hpp
```diff
@@ -19,22 +20,25 @@ struct BtcBlockAddon {
   //! reference counter for fork resolution
   uint32_t refCounter = 0;
 
+  bool operator==(const BtcBlockAddon& o) const {
+    bool a = refCounter == o.refCounter;
+    bool b = chainWork == o.chainWork;
+    return a && b;
```
blackyblack, line 26: Same here.

include/veriblock/storage/repo_batch_adaptor.hpp
```diff
@@ -0,0 +1,51 @@
...
+    pop_.saveTip(value);
+    return true;
+  }
+  bool writeTip(const BlockIndex<VbkBlock>& value) override {
+    pop_.saveTip(value);
+    return true;
+  }
+  bool writeTip(const BlockIndex<AltBlock>& value) override {
+    pop_.saveTip(value);
+    return true;
```
blackyblack, line 42: It doesn't look like a batch at all. Is it some kind of a temporary class?

include/veriblock/blockchain/base_block_tree.hpp
Warchant: compiler will optimize this away, but at the same time this layout simplifies debugging

include/veriblock/storage/repo_batch_adaptor.hpp
Warchant: It is not temporary. We use `flush` to actually write on disk, right? If so, it should not matter whether we use batch or no

blackyblack: I do not see how the inheritance from the Batch class is being used here. All I see is a collection of PopStorage method wrappers. You could simply use PopStorage methods directly and remove this helper class completely.

PR #294, opened 2020-07-14
Update mempool

Mr-Leshiy:


src/mempool.cpp
```diff
@@ -127,26 +124,35 @@ bool MemPool::submit(const ATV& atv,
                      duplicate->toShortPrettyString()));
   }
 
+  std::vector<uint8_t> endorsed_hash =
+      tree.getParams().getHash(atv.transaction.publicationData.header);
+  auto* endorsed_index = tree.getBlockIndex(endorsed_hash);
+
+  if (endorsed_index != nullptr &&
+      atv.containingBlock.height - endorsed_index->height > window) {
```
Warchant, line 132: please rename `containingBlock` to `blockOfProof` as it is a VBK block which contains a proof.
Also, ATV is in mempool, so it has no containing block yet. Get current tip height - window + 1 as max endorsed height

Mr-Leshiy: fixed.

PR #297, opened 2020-07-16
Payloads Storage

Warchant:


src/storage/payloads_storage.cpp
```diff
@@ -3,35 +3,167 @@
...
+
+  return it->second;
+}
+
+Repository& PayloadsStorage::getRepo() { return *repo_; }
+
+void PayloadsStorage::setValidity(Slice<const uint8_t> containingBlock,
+                                  Slice<const uint8_t> payloadId,
+                                  bool validity) {
+  auto key = makeKey(containingBlock, payloadId);
```
blackyblack, line 29: Please do something with this "makeKey" name. It looks awful.

src/storage/payloads_storage.cpp
```diff
@@ -3,35 +3,167 @@
...
+
+  // do nothing. any entry that is not in this map is valid by default
 }
-template <>
-const PayloadsRepository<ATV>& PayloadsStorage::getRepo() const {
-  return *_repoAtv;
+
+const std::set<AltBlock::hash_t>& PayloadsStorage::getContainingAltBlocks(
+    const std::vector<uint8_t>& payloadId) {
+  return payload_in_alt[payloadId];
```
blackyblack, line 47: I do not like that we create empty records every time we call this method.

include/veriblock/entities/altblock.hpp
```diff
@@ -123,13 +124,6 @@ inline void PrintTo(const AltBlock& block, ::std::ostream* os) {
...
-template <>
-struct hash<std::vector<uint8_t>> {
-  size_t operator()(const std::vector<uint8_t>& x) const {
-    return std::hash<std::string>{}(std::string{x.begin(), x.end()});
-  }
-};
-
 template <>
 struct hash<altintegration::AltBlock> {
   size_t operator()(const altintegration::AltBlock& block) {
```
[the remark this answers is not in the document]
Warchant: No. Hashers of our entities can and should reside inside same hpp with an entity.

std::hash<std::vector<uint8_t>> is an exception, because std::vector is std type

src/storage/payloads_storage.cpp
Warchant: this is intentional. otherwise, we would have to return a copy of set

blackyblack: What's wrong with having a copy of set? Do you want to modify it after calling get? I think it should be modifiable via some method. And you have a const ref anyway.

PR #310, opened 2020-07-24
fix overzealous VTB/ATV deduplication, improve performance by deduplicating VTBs/ATVs in AltTree

eu321:


src/entities/vtb.cpp
```diff
@@ -50,9 +50,11 @@ std::vector<uint8_t> VTB::toVbkEncoding() const {
...
 
 VTB::id_t VTB::getId() const {
-  auto left = transaction.bitcoinTransaction.getHash();
-  auto right = transaction.blockOfProof.getHash();
-  return sha256(left, right);
+  auto btcTx = transaction.bitcoinTransaction.getHash();
+  auto blockOfProof = transaction.blockOfProof.getHash();
+  auto containingVbkBlock = uint256(containingBlock.getHash());
+  auto temp = sha256(blockOfProof, containingVbkBlock);
+  return sha256(btcTx, temp);
```
Warchant, line 57: This opens opportunity for spam. Attacker can create many VBK blocks with **existing** VTB (which already resides somewhere on the main chain), and all altchains will take this VTB, because they will have different IDs.

Better option would be to have 2 ways to calculate ID.
1. GetID(VTB) in altchain - this algo
2. GetID(VTB) in vbkchain - old algo

Old algo will be used to search for duplicates (to prevent issue I described above) in VBK tree, new algo will be used to search for duplicates in ALT tree.

Warchant: We can do conditional compile-time selection of algorithm:
```
template <typename Tag>
id_t GetId(const VTB&);

template <>
id_t GetId<struct AltTag>(const VTB& vtb) {...}

template <>
id_t GetId<struct VbkTag>(const VTB& vtb) {...}
```

Then in AltTree use `GetId<AltTag>(vtb);`, and in VbkTree use `GetId<VbkTag>(vtb);`

eu321: The old algo catches everything that the new one does.

Reuse of BTC endorsements is a valid scenario. Technically, reusing an endorsement renders a VBK chain invalid.

We discussed it with Max before:
* to completely validate VBK blocks(NOT block headers) it is necessary to have complete block contents.
* the alt chain has only a limited view of a small subset of transactions
* thus, it is to be expected that the altchain unknowingly operates on VBK blocks and chains that are invalid and PoP algos have to be resilient to that
* it is sometimes possible to prove that a VBK block is invalid using the limited information available to the altchain, but the altchain is not required to take this into consideration.

As to spam, I see the following scenarios:
* mempool spam: an early validation algo would delete reused endorsements as strictly inferior to the original ones =  no impact
* honest regular pow miners: would assemble blocks from mempool contents, have an incentive to prioritize useful pop data.
* forks far in the past(potentially easier to make due to retargetting rules): should be addressed by orphan fork detection code
* recent forks:  it's just too expensive -- mining an alternative chain is not worth it just to add a shallow spammy candidate fork.

The old algo even after being fixed(see the ATV PR)  prevents 2 altchain forks from being simultaneously applied and violates a very useful invariant in the VBK tree: all blocks are valid, addPayload refuses to add VTBs that invalidate the containing block, adding a VTB doesn't invalidate any other blocks. This has correctness, performance and complexity implications.

include/veriblock/blockchain/blockchain_util.hpp
```diff
@@ -83,8 +83,10 @@ bool recoverEndorsedBy(ProtectedBlockTree& ed_,
     // validation, no 'action' is actually executed.
     actions.push_back([&] {
       auto& by = endorsed->endorsedBy;
-      VBK_ASSERT(std::find(by.rbegin(), by.rend(), &e) == by.rend() &&
-                 "same endorsement is added to endorsedBy second time");
+      if (!ProtectedBlockTree::index_t::addEndorsementAllowDuplicates) {
+        VBK_ASSERT(std::find(by.rbegin(), by.rend(), &e) == by.rend() &&
+                   "same endorsement is added to endorsedBy second time");
+      }
```
Warchant, line 89: here you can leave this assert, because it looks for duplicates by address, not by id

src/blockchain/alt_block_tree.cpp
```diff
@@ -142,6 +142,45 @@ bool AltTree::addPayloads(const AltBlock::hash_t& containing,
...
+        tree.getStorage().getContainingAltBlocks(vtb.getId().asVector());
+    for (const auto& containingBlock : containingBlocks) {
+      auto* containingIndex = tree.getBlockIndex(containingBlock);
+      VBK_ASSERT(
+          containingIndex != nullptr &&
+          "state corruption: the storage index and block tree are out of sync");
+
+      if (containingIndex->getHeight() > index.getHeight()) {
+        if (containingIndex->getAncestor(index.getHeight()) == &index) {
+          // TODO: flag the descendant block and its VTB copy as invalid
```
Warchant, line 161: you can safely add storage_.setValidity(..., false);

PR #315, opened 2020-07-28
update submit for the VBK block

Mr-Leshiy:


src/mempool.cpp
```diff
@@ -359,13 +357,11 @@ bool MemPool::submit(const VbkBlock& blk, ValidationState& state) {
   }
 
   // stateful validation
-  if (tree_->vbk().getBlockIndex(blk.getHash())) {
+  if (!tree_->vbk().getBlockIndex(blk.getHash())) {
     // duplicate
-    return state.Invalid("pop-mempool-submit-vbkblock-stateful");
+    touchVbkBlock(blk, blk.getId());
   }
```
blackyblack, line 363: I'd inverse it:
if (tree_->vbk().getBlockIndex(blk.getHash())) {
  return true;
}

PR #326, opened 2020-08-07
Add BlockOfProof vector to block index

Warchant:
For BFI we need to quickly get a list of endorsements which have same (given) blockOfProof


include/veriblock/entities/btcblock.hpp
```diff
@@ -113,6 +111,9 @@ struct BtcBlock {
   uint32_t timestamp = 0;
   uint32_t bits = 0;
   uint32_t nonce = 0;
+
+ private:
+  static const std::string _name;
```
blackyblack, line 116: I think it is safe to have it public.

Warchant: Yes and no. We should use only `::name()` static method to get this field

include/veriblock/blockchain/btc_block_addon.hpp
```diff
@@ -19,6 +20,10 @@ struct BtcBlockAddon {
   //! block
   ArithUint256 chainWork = 0;
 
+  //! (memory-only) a list of endorsements of VBK blocks, whose BlockOfProof is
+  //! this block. must be a vector, because we can have duplicates here
+  std::vector<VbkEndorsement*> blockOfProofEndorsements;
```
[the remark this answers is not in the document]
Warchant: This is part of state, meaning that it is similar to `endorsedBy` vector, unfortunately it can't :(

PR #335, opened 2020-08-13
Add BLOCK_HAS_PAYLOADS flag

Warchant:


Warchant:
Breaking, because adds `status` to block serialization.

src/blockchain/alt_block_tree.cpp
```diff
@@ -46,259 +46,91 @@ bool AltTree::bootstrap(ValidationState& state) {
...
-      if (containingIndex == nullptr) {
-        return state.Invalid(
-            "ALT-duplicate-payloads-" + P::name() + "-ancestor",
-            fmt::sprintf(
-                "Ancestor block does not exist in the chain %s that we "
-                "attempted to add to block=%s.",
-                P::name(),
-                index.toPrettyString()));
+template <typename Element, typename Container>
+auto find_duplicates(BlockIndex<AltBlock>& index, Container& pop, AltTree& tree)
```
blackyblack, line 64: findDuplicates

blackyblack: I think index can be const here.

src/blockchain/alt_block_tree.cpp
```diff
@@ -46,259 +46,91 @@ bool AltTree::bootstrap(ValidationState& state) {
...
+    -> decltype(pop.end()) {
+  const auto startHeight = tree.getParams().getBootstrapBlock().height;
+  Chain<BlockIndex<AltBlock>> chain(startHeight, &index);
+  std::unordered_set<std::vector<uint8_t>> ids;
+
+  const auto& storage = tree.getStorage();
+  auto newend = std::remove_if(pop.begin(), pop.end(), [&](const Element& p) {
```
blackyblack, line 71: Pop is being modified so the function should be name smth like findDuplicatesAndFilterPop

src/blockchain/alt_block_tree.cpp
```diff
@@ -46,259 +46,91 @@ bool AltTree::bootstrap(ValidationState& state) {
...
-    }    // end for
-
-    it = failedPayload ? payloads.erase(it) : ++it;
-  }  // end for
+template <typename Pop>
+bool searchForDuplicates(BlockIndex<AltBlock>& index,
+                         std::vector<Pop>& pop,
+                         AltTree& tree,
+                         ValidationState& state,
+                         bool continueOnInvalid) {
```
blackyblack, line 95: It is not just duplicates search but also Pop filtering. Are you sure you want to modify pop structure?

Warchant: Yes. It is used in mempool.

PR #336, opened 2020-08-14
Feature/add context asserts

Mr-Leshiy:


src/blockchain/alt_block_tree.cpp
```diff
@@ -299,6 +298,16 @@ bool checkNoPayloadDuplicatesInOtherBlocks(AltTree& tree,
   return true;
 }
 
+template <typename pop_t>
+void payloadsContextEmptyAssert(const std::vector<pop_t>& payloads) {
```
Warchant, line 302:
```suggestion
void assertContextEmpty(const std::vector<pop_t>& payloads) {
```

Warchant: please always start such funcs with word `assert`

Mr-Leshiy: fixed

PR #338, opened 2020-08-14
Feature/remove context

Mr-Leshiy:


pypopminer/src/miner.cpp
```diff
@@ -25,9 +25,6 @@ struct Payloads {
     list listctx;
```
Warchant, line 25: you should add `std::vector<VbkBlock> context;` in this struct

and serialize each block and then `.append` in this list

PR #341, opened 2020-08-17
implement OrphanBlockContainer

eu321:


include/veriblock/blockchain/orphan_block_container.hpp
```diff
@@ -0,0 +1,243 @@
...
+    return true;
+  }
+
+  /**
+   * Check if a given block is an orphan (cannot be added to the underlying tree
+   * temporarily)
+   * @return true if the block is an orphan
+   */
+  bool isOrphan(const block_t& block) const {
+    return !_tree.isBlockInitialized(block.previousBlock);
```
Warchant, line 129: I think you should inline this func here

PR #343, opened 2020-08-18
VbkTree: handle chained payloads in addPayloads() correctly

eu321:


Warchant:
Please create similar PR with base `release/beta`

PR #349, opened 2020-08-24
add vbk context validation

Mr-Leshiy:


src/mempool.cpp
```diff
@@ -446,4 +451,18 @@ bool MemPool::checkContextually<ATV>(const ATV& atv, ValidationState& state) {
   return true;
 }
 
+template <>
+bool MemPool::checkContextually<VbkBlock>(const VbkBlock& blk,
+                                          ValidationState& state) {
+  if (tree_->vbk().getBlockIndex(blk.previousBlock) == nullptr &&
+      get<VbkBlock>(blk.previousBlock) == nullptr) {
+    return state.Invalid(
+        "potential-stale-vbk-block",
```
Warchant, line 460:
```suggestion
        "bad-prev",
```

src/mempool.cpp
```diff
@@ -446,4 +451,18 @@ bool MemPool::checkContextually<ATV>(const ATV& atv, ValidationState& state) {
...
+        fmt::sprintf("Block=%s does not connect to the known chains",
```
Warchant, line 461:
```suggestion
        fmt::sprintf("Block=%s does not connect to known VBK tree",
```

Warchant (approved):
Make these fixes and then can be merged

src/mempool.cpp
Mr-Leshiy: fixed

src/mempool.cpp
Mr-Leshiy: fixed

PR #355, opened 2020-08-26
feat!(AltTree): add sync and async block validation, BLOCK_CONNECTED flag and connectedness enforcement

eu321:
…

### DIFF
--- a/include/veriblock/blockchain/block_status.hpp
+++ b/include/veriblock/blockchain/block_status.hpp
@@ -1,0 +1,51 @@
+// Copyright (c) 2019-2020 Xenios SEZC
+// https://www.veriblock.org
+// Distributed under the MIT software license, see the accompanying
+// file LICENSE or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef ALT_INTEGRATION_INCLUDE_VERIBLOCK_BLOCKCHAIN_BLOCK_STATUS_HPP_
+#define ALT_INTEGRATION_INCLUDE_VERIBLOCK_BLOCKCHAIN_BLOCK_STATUS_HPP_
+
+namespace altintegration {
+
+enum BlockStatus : uint32_t {
+  //! default state for validity - validity state is unknown
+  BLOCK_VALID_UNKNOWN = 0,
+  //! acceptBlock succeded. All ancestors are at least at this state.
+  BLOCK_VALID_TREE = 1 << 0,
+  //! all validity flags
+  BLOCK_VALID_MASK = BLOCK_VALID_TREE,
+  //! this is a bootstrap block
+  BLOCK_BOOTSTRAP = 1 << 1,
+  //! block is statelessly valid, but the altchain marked it as failed
+  BLOCK_FAILED_BLOCK = 1 << 2,
+  //! block failed state{less,ful} validation due to its payloads
+  BLOCK_FAILED_POP = 1 << 3,
+  //! block is state{lessly,fully} valid and the altchain did not report it as
+  //! invalid, but some of the ancestor blocks are invalid
+  BLOCK_FAILED_CHILD = 1 << 4,
+  //! all invalidity flags
+  BLOCK_FAILED_MASK =
+      BLOCK_FAILED_CHILD | BLOCK_FAILED_POP | BLOCK_FAILED_BLOCK,
+  //! the block has been applied via PopStateMachine
+  BLOCK_APPLIED = 1 << 5,
+  //! the block has been at least once successful applied via PopStateMachine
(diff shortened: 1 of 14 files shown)


include/veriblock/blockchain/block_status.hpp
```diff
@@ -0,0 +1,51 @@
...
+  BLOCK_FAILED_MASK =
+      BLOCK_FAILED_CHILD | BLOCK_FAILED_POP | BLOCK_FAILED_BLOCK,
+  //! the block has been applied via PopStateMachine
+  BLOCK_APPLIED = 1 << 5,
+  //! the block has been at least once successful applied via PopStateMachine
+  BLOCK_CAN_BE_APPLIED = 1 << 6,
+  //! if set, acceptBlock has been executed on this block
+  BLOCK_HAS_PAYLOADS = 1 << 7,
+  //! the block is connected via connectBlock
+  BLOCK_CONNECTED = 1 << 8,
```
Warchant, line 37: how about 
```
BLOCK_VALID_HEADER = 1, // BLOCK_VALID_TREE
BLOCK_HAS_PAYLOADS = 2,
BLOCK_CONNECTED = 3, 
BLOCK_VALID_POP =4 // CAN_BE_APPLIED
```

then use `raiseValidity` instead of setting all flags. 
and use `isValid(level)` to check if a block met certain conditions.

src/blockchain/alt_block_tree.cpp
```diff
@@ -75,105 +77,200 @@ void commitPayloadsIds(BlockIndex<AltBlock>& index,
...
+    VBK_ASSERT_MSG(index.allDescendantsUnconnected(),
+                   "a descendant of block %s is connected",
+                   index.toPrettyString());
+  }
+
+  // partial stateful validation
+  // FIXME: eventually we want to perform full stateful validation here,
+  // effectively find out whether setState(index) will be successful
+
+  // BUG: applyBlock will clear BLOCK_FAILED_POP flag since it does not check
```
Warchant, line 235: could you elaborate?

src/blockchain/alt_block_tree.cpp
```diff
@@ -75,105 +77,200 @@ void commitPayloadsIds(BlockIndex<AltBlock>& index,
...
-                         fmt::format("Found {} duplicated {}",
-                                     std::distance(newend, pop.end()),
-                                     Pop::name()));
-  }
+template <typename P, typename T>
+bool hasDuplicates(BlockIndex<AltBlock>& index,
+                   const std::vector<T>& payloads_ref,
+                   AltTree& tree,
+                   ValidationState& state) {
+  std::vector<T> payloads = payloads_ref;
```
Warchant, line 140: you can pass payloads by value here

src/blockchain/alt_block_tree.cpp
```diff
@@ -75,105 +77,200 @@ void commitPayloadsIds(BlockIndex<AltBlock>& index,
...
+      ValidationState dummy;
+      connectBlock(*successor, dummy);
+    }
+  }
+
+  return index.isValid();
+}
+
+// in !StrictAddPayloadsOrdering mode, payloads can be added to any block, which
+// may trigger incorrect AltTree behavior in certain cases
```
Warchant, line 265: then we should disable !StrictAddPayloadsOrdering?

include/veriblock/blockchain/block_status.hpp
eu321: It might make sense to go this way, but first we need to decide on the list off flags/validation levels. As it stands, full validation happens in multiple successive steps and what we seem to need is to record the last successful step and the last unsuccessful one( to be able to differentiate between a failure and not having attempted the next step)

src/blockchain/alt_block_tree.cpp
eu321: PopStateMachine.applyBlock() will happily attempt to apply a BLOCK_FAILED_POP block and clear the flag if it is successful, however applyBlock() doesn't perform complete validation of blocks. AFAIK, we aren't using this feature or "feature" at this moment, but eventually we need to either get rid of it or fix it.

src/blockchain/alt_block_tree.cpp
eu321: agree

src/blockchain/alt_block_tree.cpp
eu321: We use StrictAddPayloadsOrdering as a stopgap for tests. Those certain cases are basically related to duplicate detection. We can add duplicate payloads to an ancestor block after adding them to a descendant and the duplicates will not be detected and handled properly

Warchant (approved):

PR #356, opened 2020-08-26
Add strongly equivalent VTB

Mr-Leshiy:


src/mempool_prioritization.cpp
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2019-2020 Xenios SEZC
+// https://www.veriblock.org
+// Distributed under the MIT software license, see the accompanying
+// file LICENSE or http://www.opensource.org/licenses/mit-license.php.
+
+#include "veriblock/mempool_prioritization.hpp"
+
+namespace altintegration {
+
+bool isStronglyEquivalent(const VTB& vtb1,
```
Warchant, line 10: I think since you set asserts that containing blocks must exist on VBK tree, it would be better to move this function to VbkBlockTree (as member func)

src/mempool_prioritization.cpp
```diff
@@ -0,0 +1,42 @@
...
+  } else {
+    is_on_the_same_chain =
+        vtb2_containing_block->getAncestor(
+            vtb1_containing_block->getHeight()) == vtb1_containing_block;
+  }
+
+  return (vtb1.transaction.bitcoinTransaction ==
+          vtb2.transaction.bitcoinTransaction) &&
+         (vtb1.transaction.blockOfProof == vtb1.transaction.blockOfProof) &&
+         is_on_the_same_chain;
```
Warchant, line 39: I think whether VTBs are on same chain or not should not imfluence their 'strong equality'
so technically it is better to calculate StronglyEquivalentIds for VTBs (which do not include containing block), and separately add member func to `VbkBlockTree` `bool AreOnSameChain(const VbkBlock& b1, const VbkBlock& b2) const`

Warchant (approved):

PR #363, opened 2020-08-31
Feature/applied block counting optimize fr

eu321:


include/veriblock/blockchain/pop/pop_state_machine.hpp
```diff
@@ -202,6 +203,7 @@ struct PopStateMachine {
     }
 
     index.unsetFlag(BLOCK_APPLIED);
+    --ed_.appliedBlockCount;
```
blackyblack, line 206: I'd add an assert that appliedBlockCount is not zero.

eu321: there's a more elaborate check in setState() that would also catch this case, but it doesn't hurt, yeah

include/veriblock/blockchain/base_block_tree.hpp
```diff
@@ -340,6 +342,14 @@ struct BaseBlockTree {
     return validity_sig_.disconnect(id);
   }
 
+  index_t& getRoot() {
```
Warchant, line 345: In our terms this is called "bootstrap block" 
```suggestion
  index_t& getBootstrapBlock() {
```

include/veriblock/blockchain/base_block_tree.hpp
```diff
@@ -340,6 +342,14 @@ struct BaseBlockTree {
...
+    VBK_ASSERT_MSG(isBootstrapped(), "must be bootstrapped");
+    return *getBestChain().first();
+  }
+
+  //! the number of blocks that have BLOCK_APPLIED flag set
+  block_height_t appliedBlockCount = 0;
```
Warchant, line 351: should it be public?

include/veriblock/blockchain/pop/fork_resolution.hpp
```diff
@@ -506,7 +510,7 @@ struct PopAwareForkResolutionComparator {
     {
       auto guard = ing_->deferForkResolutionGuard();
 
-      if (!sm.apply(*chainB.first(), *chainB.tip(), state, false)) {
+      if (!sm.apply(*chainB.first(), *chainB.tip(), state)) {
```
Warchant, line 513: why did you change this? 
this apply applies chain B together with chain A, and this `false` indicated that we should not set flag `CAN_BE_APPLIED` during this state change

include/veriblock/blockchain/pop/pop_state_machine.hpp
```diff
@@ -170,10 +163,18 @@ struct PopStateMachine {
                  "its payloads");
     }
 
-    index.setFlag(BLOCK_APPLIED);
-    if (shouldSetCanBeApplied) {
+    // if the applied block count equals the size of the chain between the root
```
Warchant, line 166:
```suggestion
    // if the applied block count equals the size of the chain between the bootstrap block
```

Warchant:
Please open same PR based on `release/beta` branch

include/veriblock/blockchain/base_block_tree.hpp
eu321: either it has to be public since it's accessed by other classes, or we need to add blockApplied() blockUnapplied(), getAppliedBlockCount()

include/veriblock/blockchain/pop/pop_state_machine.hpp
eu321: there can be several bootstrap blocks, while the root block refers to the root of the tree

Warchant: ok, add a comment `the root block (first bootstrap block)`

Warchant (approved):

PR #371, opened 2020-09-07
replace multiple block validity flags with a single validity level; add more asserts

eu321:
…

### DIFF
--- a/include/veriblock/blockchain/base_block_tree.hpp
+++ b/include/veriblock/blockchain/base_block_tree.hpp
@@ -134,7 +134,7 @@ struct BaseBlockTree {
       current->pprev->pnext.insert(current);
     }
 
-    current->setFlag(BLOCK_VALID_TREE);
+    current->raiseValidity(BLOCK_VALID_TREE);
     current->unsetDirty();
 
     tryAddTip(current);
@@ -639,6 +639,13 @@ struct BaseBlockTree {
   }
 
   void doInvalidate(index_t& block, enum BlockStatus reason) {
+    VBK_ASSERT_MSG(
+        !block.isValidUpTo(BLOCK_CAN_BE_APPLIED) ||
+            (reason & BLOCK_FAILED_POP) == 0u,
+        "attempted to set mutually exclusive flags BLOCK_CAN_BE_APPLIED "
+        "and BLOCK_FAILED_POPfor block %s",
+        block.toPrettyString());
+
     block.setFlag(reason);
     validity_sig_.emit(block);
   }
--- a/include/veriblock/blockchain/block_index.hpp
+++ b/include/veriblock/blockchain/block_index.hpp
@@ -43,7 +43,9 @@ struct BlockIndex : public Block::addon_t {
    * ancestors are connected.
    * @return true if block is connected, false otherwise.
    */
-  bool isConnected() const noexcept { return this->hasFlags(BLOCK_CONNECTED); }
+  bool isConnected() const noexcept {
+    return this->isValidUpTo(BLOCK_CONNECTED);
+  }
 
   uint32_t getStatus() const { return status; }
   void setStatus(uint32_t _status) {
@@ -51,13 +53,32 @@ struct BlockIndex : public Block::addon_t {
     setDirty();
   }
 
+  uint32_t getValidityLevel() const {
+    auto level = status & BLOCK_VALID_MASK;
+    VBK_ASSERT_MSG(level <= BLOCK_CAN_BE_APPLIED, "unknown validity level");
+    return level;
+  }
+
   bool isValid(enum BlockStatus upTo = BLOCK_VALID_TREE) const {
-    VBK_ASSERT(!(upTo & ~BLOCK_VALID_MASK));  // Only validity flags allowed.
     if ((status & BLOCK_FAILED_MASK) != 0u) {
       // block failed
       return false;
     }
-    return ((status & BLOCK_VALID_MASK) >= upTo);
+    return isValidUpTo(upTo);
+  }
+
+  bool isValidUpTo(enum BlockStatus upTo) const {
+    VBK_ASSERT_MSG(!(upTo & ~BLOCK_VALID_MASK),
+                   "Only validity flags are allowed");
+
+    auto validityLevel = getValidityLevel();
+    VBK_ASSERT_MSG(
+        validityLevel != BLOCK_CAN_BE_APPLIED || !hasFlags(BLOCK_FAILED_POP),
+        "block %s is both BLOCK_CAN_BE_APPLIED and BLOCK_FAILED_POP which are "
+        "mutually exclusive",
+        toPrettyString());
+
+    return getValidityLevel() >= upTo;
   }
 
   void setNull() {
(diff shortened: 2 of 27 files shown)


include/veriblock/blockchain/block_index.hpp
```diff
@@ -43,21 +43,42 @@ struct BlockIndex : public Block::addon_t {
...
+    return isValidUpTo(upTo);
+  }
+
+  bool isValidUpTo(enum BlockStatus upTo) const {
+    VBK_ASSERT_MSG(!(upTo & ~BLOCK_VALID_MASK),
+                   "Only validity flags are allowed");
+
+    auto validityLevel = getValidityLevel();
+    VBK_ASSERT_MSG(
+        validityLevel != BLOCK_CAN_BE_APPLIED || !hasFlags(BLOCK_FAILED_POP),
```
blackyblack, line 76: Here also can simplify with:
!(validityLevel == BLOCK_CAN_BE_APPLIED && hasFlags(BLOCK_FAILED_POP))
that is also consistent with the comment below

include/veriblock/blockchain/block_status.hpp
```diff
@@ -29,12 +25,23 @@ enum BlockStatus : uint32_t {
...
+  //! acceptBlock has been executed on this block; payloads are statelessly
+  //! valid
+  BLOCK_HAS_PAYLOADS = 2 << 6,
   //! the block is connected via connectBlock
-  BLOCK_CONNECTED = 1 << 8,
+  BLOCK_CONNECTED = 3 << 6,
+  //! the block has been successfully applied, likely along with another chain
+  BLOCK_HAS_BEEN_APPLIED = 4 << 6,
+  //! the chain with the block at its tip is fully valid
+  BLOCK_CAN_BE_APPLIED = 5 << 6,
```
blackyblack, line 39: I don't think you can build flags like this. For example BLOCK_VALID_TREE | BLOCK_HAS_BEEN_APPLIED equals BLOCK_CAN_BE_APPLIED.

include/veriblock/blockchain/block_status.hpp
```diff
@@ -29,12 +25,23 @@ enum BlockStatus : uint32_t {
...
+  //! acceptBlockHeader succeded. All ancestors are at least at this state.
+  BLOCK_VALID_TREE = 1 << 6,
+  //! acceptBlock has been executed on this block; payloads are statelessly
+  //! valid
+  BLOCK_HAS_PAYLOADS = 2 << 6,
   //! the block is connected via connectBlock
-  BLOCK_CONNECTED = 1 << 8,
+  BLOCK_CONNECTED = 3 << 6,
+  //! the block has been successfully applied, likely along with another chain
+  BLOCK_HAS_BEEN_APPLIED = 4 << 6,
```
blackyblack, line 37: What's the difference between BLOCK_HAS_BEEN_APPLIED and BLOCK_APPLIED?

include/veriblock/blockchain/pop/pop_state_machine.hpp
```diff
@@ -154,23 +151,17 @@ struct PopStateMachine {
...
     // so it is fully valid
-    if (index.pprev->hasFlags(BLOCK_CAN_BE_APPLIED) &&
+    if (index.pprev->isValid(BLOCK_CAN_BE_APPLIED) &&
         index.getHeight() ==
             ed_.getRoot().getHeight() + ed_.appliedBlockCount) {
-      index.setFlag(BLOCK_CAN_BE_APPLIED);
+      index.raiseValidity(BLOCK_CAN_BE_APPLIED);
+    } else {
+      index.raiseValidity(BLOCK_HAS_BEEN_APPLIED);
     }
```
blackyblack, line 165: This else statement looks weird. So if ed_.appliedBlockCount is less than index.height block will get BLOCK_HAS_BEEN_APPLIED status?

src/blockchain/alt_block_tree.cpp
```diff
@@ -173,13 +172,10 @@ void AltBlockTree::setPayloads(index_t& index, const PopData& payloads) {
                payloads.atvs.size(),
                index.toShortPrettyString());
 
-  VBK_ASSERT_MSG(!index.hasFlags(BLOCK_HAS_PAYLOADS),
+  VBK_ASSERT_MSG(index.getValidityLevel() == BLOCK_VALID_TREE,
                  "block %s already contains payloads",
```
blackyblack, line 176: Assert message is wrong

include/veriblock/blockchain/block_status.hpp
eu321: BLOCK_VALID_TREE and BLOCK_CAN_BE_APPLIED aren't flags as such,  they are mutually exclusive. Effectively we have a 3-bit uint here.

include/veriblock/blockchain/block_status.hpp
eu321: naming of BLOCK_CAN_BE_APPLIED is also awful. We need to rename them. BLOCK_APPLIED means it is applied right now. BLOCK_HAS_BEEN_APPLIED is a validity status - has been applied before

include/veriblock/blockchain/pop/pop_state_machine.hpp
eu321: not sure I understand the question. this functionality isn't changed by this PR

src/blockchain/alt_block_tree.cpp
eu321: it may not be technically correct as in theory it could be called on a block with BLOCK_VALID_UNKNOWN status due to a bug, but in all other cases setPayloads() is the function that raises the validity level from  BLOCK_VALID_TREE to BLOCK_HAS_PAYLOADS

include/veriblock/blockchain/block_status.hpp
eu321: thinking more about it, maybe we should use a proper bit field here. will discuss it with Bohdan

blackyblack: Yes I think they should be separate from flags to avoid confusion.

include/veriblock/blockchain/base_block_tree.hpp
```diff
@@ -639,6 +639,13 @@ struct BaseBlockTree {
   }
 
   void doInvalidate(index_t& block, enum BlockStatus reason) {
+    VBK_ASSERT_MSG(
+        !block.isValidUpTo(BLOCK_CAN_BE_APPLIED) ||
+            (reason & BLOCK_FAILED_POP) == 0u,
+        "attempted to set mutually exclusive flags BLOCK_CAN_BE_APPLIED "
+        "and BLOCK_FAILED_POPfor block %s",
```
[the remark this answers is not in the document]
eu321: maybe you're right. I'm used to transformations like !(A &&B) <=> !A ||!B and the latter form is easier to read for me

include/veriblock/blockchain/block_status.hpp
Warchant: maybe name it `BLOCK_VALID`?

test/e2e/scenario_2_test.cpp
```diff
@@ -89,7 +89,7 @@ TEST_F(Scenario2, scenario_2) {
                  vtbs[0].containingBlock.getHash(),
                  popminer->vbk());
 
-  // Step 1
+  VBK_LOG_DEBUG("Step 1");
```
Warchant, line 92: remove these steps

include/veriblock/blockchain/pop/pop_state_machine.hpp
blackyblack: Ok I see that BLOCK_CAN_BE_APPLIED is higher validity than BLOCK_HAS_BEEN_APPLIED. So the logic here is to set BLOCK_HAS_BEEN_APPLIED first and set BLOCK_CAN_BE_APPLIED if the previous chain is fully applied.
The logic seems OK but was hard to embrace cause is looks like BLOCK_CAN_BE_APPLIED should come before BLOCK_HAS_BEEN_APPLIED.

Warchant (approved):

PR #375, opened 2020-09-08
Feature/optimization rewards cache btc 651

blackyblack:


include/veriblock/rewards/poprewards_cache.hpp
```diff
@@ -0,0 +1,80 @@
...
+   * @param lastBlock erase cache records up to this block, not including.
+   */
+  void cacheTruncateTailTo(const index_t& lastBlock);
+
+  /**
+   * Update cache to have scores of this block and any additional blocks
+   * required to calculate POP rewards.
+   * @param endorsed reward will be paid for this block.
+   */
+  void cacheRebuild(const index_t& endorsed);
```
eu321, line 75: the API doc provides a hint as to what name would be more descriptive: updateCache ;)

include/veriblock/rewards/poprewards_cache.hpp
```diff
@@ -0,0 +1,80 @@
...
+  const VbkBlockTree* vbkTree_;
+  ring_buffer<pair_t> buffer;
+
+  PopRewardsBigDecimal updateAndCalculateDifficulty(const index_t& tip);
+
+  void cacheAddBlock(const index_t& block);
+
+  bool cacheBlockExists(const index_t* block);
+
+  void cacheInvalidate(const index_t& endorsed);
```
eu321, line 60: This function doesn't just invalidate the difficulty score for the endorsed block(and its descendants), but rebuilds the cache completely. Maybe rebuildCache would be a better name?

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+  return calculateDifficulty(endorsed);
+}
+
+void PopRewardsCache::cacheAddBlock(const index_t& block) {
+  auto score = scoreFromEndorsements(block);
+  buffer.push_back({block, score});
+}
+
+bool PopRewardsCache::cacheBlockExists(const index_t* block) {
+  if (block == nullptr) {
```
eu321, line 61: just pass const index_t& block

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+  if (difficulty < 1.0) {
+    difficulty = 1.0;
+  }
+  return difficulty;
+}
+
+std::map<std::vector<uint8_t>, int64_t> PopRewardsCache::calculatePayouts(
+    const index_t& endorsedBlock) {
+  auto popDifficulty = updateAndCalculateDifficulty(endorsedBlock);
+  auto it = std::find_if(
```
eu321, line 40: duplicates cacheBlockExists code

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+// file LICENSE or http://www.opensource.org/licenses/mit-license.php.
+
+#include <veriblock/rewards/poprewards_cache.hpp>
+
+namespace altintegration {
+
+PopRewardsBigDecimal PopRewardsCache::calculateDifficulty(
+    const index_t& /* ignore */) const {
+  auto it = buffer.crbegin();
+  // remove last element - it is the current block that we calculate difficulty
```
eu321, line 13: such a bold statement definitely needs an assert()

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+      });
+  if (it == buffer.rend()) {
+    return false;
+  }
+  return true;
+}
+
+void PopRewardsCache::cacheInvalidate(const index_t& endorsed) {
+  buffer.clear();
+  std::vector<index_t> blocks;
```
eu321, line 76: make sense to preallocate altParams_->getRewardParams().difficultyAveragingInterval() elements

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  for (const auto& b : reverse_iterate(blocks)) {
+    cacheAddBlock(b);
+  }
+}
+
+void PopRewardsCache::cacheTruncateTailTo(const index_t& lastBlock) {
+  auto it = buffer.rbegin();
+  while (true) {
```
eu321, line 93: while(!buffer.empty() && buffer.back().getHash() != lastBlock.getHash()) {
    buffer.pop_back();
}

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+    if (buffer.empty()) break;
+    if (it == buffer.rend()) break;
+    if (it->first.getHash() == lastBlock.getHash()) return;
+    buffer.pop_back();
+    it = buffer.rbegin();
+  }
+  return;
```
eu321, line 100: } on the next line does the same thing

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+void PopRewardsCache::cacheRebuild(const index_t& endorsed) {
+  auto* begin = endorsed.getAncestorBlocksBehind(
+      altParams_->getRewardParams().difficultyAveragingInterval() + 1);
+  auto* end = endorsed.pprev;
+
+  bool rangeExists = true;
+  rangeExists = rangeExists && cacheBlockExists(begin);
+  rangeExists = rangeExists && cacheBlockExists(end);
```
eu321, line 110: bool rangeExists =  cacheBlockExists(begin)&& cacheBlockExists(end);

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+  if (!rangeExists) {
```
eu321, line 112: This does not handle reasonable use cases efficiently, such as calculating scores out of order, calculating scores of forks.

Assuming we don't change the approach of storing only the current chain in the cache, an efficient implementation would seem to be:
* find the fork point between the tip of the cache and 'endorsed'
* drop the range between the fork and tip
* add the missing scores between the fork and 'endorsed'; the ring buffer will drop earlier blocks as necessary

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+    cacheInvalidate(endorsed);
+    return;
+  }
+
+  if (cacheBlockExists(&endorsed)) {
```
eu321, line 117: I understand that the way pop scores are currently calculated in vBTC don't hit this limitation, but nevertheless, supporting calculation of scores out of order should not be that hard

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+  buffer.push_back({block, score});
+}
+
+bool PopRewardsCache::cacheBlockExists(const index_t* block) {
+  if (block == nullptr) {
+    return false;
+  }
+  auto it =
+      std::find_if(buffer.rbegin(), buffer.rend(), [block](const pair_t& item) {
+        return item.first.getHash() == block->getHash();
```
eu321, line 66: no need to compare hashes, pointers are sufficient. adding an assert to make sure the BlockIndex we found has the same hash is an option as well

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+  auto score = scoreFromEndorsements(block);
+  buffer.push_back({block, score});
+}
+
+bool PopRewardsCache::cacheBlockExists(const index_t* block) {
+  if (block == nullptr) {
+    return false;
+  }
+  auto it =
+      std::find_if(buffer.rbegin(), buffer.rend(), [block](const pair_t& item) {
```
eu321, line 65: since the ring buffer stores a chain, it is possible to look up blocks efficiently by height: the cache item we're looking for should be buffer.back().height - block.height elements away from the buffer end

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+  auto difficulty = sumscore / difficultyInterval;
+
+  // Minimum difficulty
+  if (difficulty < 1.0) {
+    difficulty = 1.0;
+  }
+  return difficulty;
+}
+
+std::map<std::vector<uint8_t>, int64_t> PopRewardsCache::calculatePayouts(
```
eu321, line 37: it would probably be easier to overload just 1 function, scoreFromEndorsements, as that's what we are caching and hide all cache handling code there

src/rewards/poprewards_cache.cpp
```diff
@@ -0,0 +1,126 @@
...
+    const index_t& /* ignore */) const {
+  auto it = buffer.crbegin();
+  // remove last element - it is the current block that we calculate difficulty
+  // for
+  ++it;
+  PopRewardsBigDecimal sumscore = 0.0;
+  size_t count = 0;
+  uint64_t difficultyInterval =
+      altParams_->getRewardParams().difficultyAveragingInterval();
+  for (; it != buffer.crend(); ++it) {
```
eu321, line 20: need a check that we are summing no less than difficultyAveragingInterval scores unless we're close to the genesis/bootstrap block

eu321, line 20: need a check that we are summing no less than difficultyAveragingInterval scores unless we're close to the genesis/bootstrap block

src/rewards/poprewards_cache.cpp
blackyblack: Nah, we won't be able to detect reorgs this way. For example we calculate score for block 90 and save it to cache. Next time we try to reuse the cache but we had a reorg and now block 90 has different score.
Therefore we rely on calculatePayouts() call that gives us a signal about currently processed block.

eu321: It is better to trigger cache invalidation explicitly rather than assume that calculatePayouts now(in the cache) means invalidateCache + calculatePayouts

Warchant (approved):

PR #376, opened 2020-09-08
Feature/assorted cleanups

eu321:


src/blockchain/alt_block_tree.cpp
```diff
@@ -183,16 +168,14 @@ void AltBlockTree::setPayloads(index_t& index, const PopData& payloads) {
...
-                 "block %s must not contain duplicate VBKs",
-                 index.toPrettyString());
-  VBK_ASSERT_MSG(!hasDuplicateIds(payloads.atvs),
-                 "block %s must not contain duplicate ATVs",
-                 index.toPrettyString());
+  // FIXME: it should be impossible to construct popData without running this
+  // check
+  ValidationState state;
+  VBK_ASSERT_MSG(
+      checkPopData(payloads, state),
```
Warchant, line 175: this check is expensive, we should not double-check that PopData is statelessly valid

eu321: This is a refactoring commit, so I didn't change any functionality. The check that was there was moved to an external function that the client code can now call. I'm not sure there was a duplicate id check before on the user side

Warchant (approved):

PR #377, opened 2020-09-08
update submit method

Mr-Leshiy:


include/veriblock/mempool.hpp
```diff
@@ -67,6 +68,36 @@ struct MemPool {
...
+   * @param[in] shouldDoContextualCheck if true, mempool performs contextual
+   * validation
+   * @return true if payload is valid, false otherwise
+   * @ingroup api
+   */
+  template <typename T>
+  bool submit(Slice<const uint8_t> bytes, ValidationState& state) {
+    ReadStream stream(bytes);
+    try {
+      T payload = T::fromVbkEncoding(stream);
```
Warchant, line 94: use Deserialize function

Warchant: https://github.com/VeriBlock/alt-integration-cpp/blob/master/src/entities/atv.cpp#L64

Mr-Leshiy: fixed

Warchant (approved):

PR #400, opened 2020-10-01
Add golang VBK Entities

enfipy:


bindings/go/entities/endorsement.go
```diff
@@ -0,0 +1,4 @@
+package entities
+
+// Endorsement ...
+type Endorsement struct{}
```
Warchant, line 4: Can be removed

bindings/go/serde.go
```diff
@@ -0,0 +1,226 @@
...
+	"encoding/hex"
+	"errors"
+	"io"
+	"math"
+	"reflect"
+	"unsafe"
+)
+
+// Parse - parses hex string to byte array.
+func Parse(src string) []byte {
```
Warchant, line 15: I do not think you need to have this func, because you already have this `hex.DecodeString` func.
It can fail if input string has non-hex chars

Warchant (approved):

PR #425, opened 2020-11-11
Feature/multithreaded pop validation gamma

blackyblack:


src/stateless_validation.cpp
```diff
@@ -414,4 +414,42 @@ bool checkPopData(const PopData& popData, ValidationState& state) {
...
+    results.push_back(validator.addCheck(vtb));
+  }
+  for (const auto& atv : popData.atvs) {
+    results.push_back(validator.addCheck(atv));
+  }
+
+  size_t index = 0;
+  for (auto& r : results) {
+    auto result = r.get();
+    if (result.IsInvalid()) {
```
Warchant, line 438:
```suggestion
    if (!result.IsValid()) {
```
because there's also `IsError()`

src/stateless_validation.cpp
```diff
@@ -414,4 +414,42 @@ bool checkPopData(const PopData& popData, ValidationState& state) {
...
+  }
+  for (const auto& atv : popData.atvs) {
+    results.push_back(validator.addCheck(atv));
+  }
+
+  size_t index = 0;
+  for (auto& r : results) {
+    auto result = r.get();
+    if (!result.IsValid()) {
+      state = result;
```
Warchant, line 439: I think this should be enough
```
return state.Invalid("popdata");
```

Warchant (approved):

PR #449, opened 2020-11-23
Fix golang IDs

enfipy:
Fix IDs for alt, vtb, vbk, btc

### DIFF
--- a/bindings/go/api/extern.go
+++ b/bindings/go/api/extern.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	entities "github.com/VeriBlock/alt-integration-cpp/bindings/go/entities"
 	ffi "github.com/VeriBlock/alt-integration-cpp/bindings/go/ffi"
 )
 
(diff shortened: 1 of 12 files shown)


bindings/go/api/extern.go
```diff
@@ -20,16 +21,28 @@ func SetOnGetBlockHeaderHash(fn func(toBeHashed []byte) []byte) {
 }
 
 // SetOnGetAtv ...
-func SetOnGetAtv(fn func(id []byte) []byte) {
-	ffi.OnGetAtv = fn
+func SetOnGetAtv(fn func(idBytes entities.AtvID) []byte) {
+	ffi.OnGetAtv = func(id []byte) []byte {
+		var idBytes entities.AtvID
+		copy(idBytes[:], id)
+		return fn(idBytes)
```
Warchant, line 28: https://stackoverflow.com/questions/28886616/convert-array-to-slice-in-go

bindings/go/api/extern.go
```diff
@@ -20,16 +21,28 @@ func SetOnGetBlockHeaderHash(fn func(toBeHashed []byte) []byte) {
 }
 
 // SetOnGetAtv ...
-func SetOnGetAtv(fn func(id []byte) []byte) {
-	ffi.OnGetAtv = fn
+func SetOnGetAtv(fn func(idBytes entities.AtvID) []byte) {
+	ffi.OnGetAtv = func(id []byte) []byte {
+		var idBytes entities.AtvID
+		copy(idBytes[:], id)
```
Warchant, line 27: do as little copies as possible

Warchant (changes requested):

bindings/go/api/extern.go
enfipy: https://stackoverflow.com/questions/30285680/how-to-convert-slice-to-fixed-size-array

bindings/go/api/extern.go
enfipy: Sadly, it's not possible here as we have different sizes for IDs. First I tried to write a similar code in `ffi/extern.go` but the whole difference was is just more boilerplate code

Warchant (approved):

PR #454, opened 2020-11-26
add contextInfoContainer calculation

Mr-Leshiy:


src/alt-util.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace altintegration {
+
+uint256 CalculateContextInfoContainerHash(const PopData& popData,
+                                          const BlockIndex<AltBlock>& prevBlock,
+                                          const AltChainParams& params) {
+  auto firstPreviousKeystone =
+      prevBlock.getAncestor(getFirstPreviousKeystoneHeight(
+          prevBlock.getHeight() + 1, params.getKeystoneInterval()));
+  auto secondPreviousKeystone =
```
Warchant, line 18: this can be nullptr if ancestor is not found (for example VBK10->getAncestor(100) or VBK10->getAncestor(-100))

Warchant: If keystone is nullptr, then make it equal to `000..000`
```
auto zeroHash = prevBlock.getHash().fill(0);
```

Warchant (changes requested):

src/alt-util.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+  auto atvMerkleRoot =
+      PayloadsMerkleTree<ATV>(map_get_id(popData.atvs)).getMerkleRoot();
+  auto vtbMerkleRoot =
+      PayloadsMerkleTree<VTB>(map_get_id(popData.vtbs)).getMerkleRoot();
+  auto vbkMerkleRoot =
+      PayloadsMerkleTree<VbkBlock>(map_get_id(popData.context)).getMerkleRoot();
+
+  uint256 left = sha256twice(atvMerkleRoot, vtbMerkleRoot);
+  uint256 right = sha256twice(vbkMerkleRoot, vbkMerkleRoot);
```
Warchant, line 41: pls change the order
```
(vbk, vtb)  (atv, context)
```

Mr-Leshiy: fixed

src/alt-util.cpp
Mr-Leshiy: fixed

PR #477, opened 2020-12-09
Fix altblock serialization

blackyblack:


src/entities/altblock.cpp
```diff
@@ -67,7 +67,7 @@ bool DeserializeFromRaw(ReadStream& stream,
   if (!readSingleByteLenValue(stream,
                               out.previousBlock,
                               state,
-                              MIN_ALT_HASH_SIZE,
+                              0,
```
Warchant, line 70: i think it would be better to change value of MIN_ALT_HASH_SIZE to 0

Warchant:
Migrated to #488

PR #482, opened 2020-12-10
Add ValidationState to the C wrapper

Mr-Leshiy:


include/veriblock/c/validation_state.h
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * @defgroup c-api C interface
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct VbkValidationState VBK_ValidationState;
```
Warchant, line 20: Can we actually *define* a normal struct here?

src/c/validation_state.cpp
```diff
@@ -7,15 +7,24 @@
...
+void VBK_FreeValidationState(VBK_ValidationState* self) {
   if (self) {
     delete self;
   }
+}
+
+const char* VBK_ValidationState_getErrorMessage(VBK_ValidationState* self) {
+  VBK_ASSERT(self);
+
+  return self->validation_state.toString().c_str();
```
Warchant, line 23: this is wrong. toString() returns temporary object and then you return a pointer to tmp object.

store std::string in VBK_ValidationState struct, then string=state.toString() , then return string.c_str()

Mr-Leshiy: fixed

Warchant (approved):

enfipy (approved):

PR #502, opened 2020-12-28
Go wrapper refactor

Mr-Leshiy:


enfipy (approved):

bindings/go/entities/popdata.go
```diff
@@ -82,35 +83,35 @@ func (v *PopData) FromVbkEncoding(stream io.Reader) error {
 	for i, context := range contexts {
 		v.Context[i] = *context.(*VbkBlock)
 	}
-	atvs, err := veriblock.ReadArrayOf(stream, 0, veriblock.MaxContextCountAltPublication, func(stream io.Reader) (interface{}, error) {
-		block := Atv{}
-		err := block.FromVbkEncoding(stream)
+	vtbs, err := veriblock.ReadArrayOf(stream, 0, veriblock.MaxContextCountVbkPublication, func(stream io.Reader) (interface{}, error) {
```
Warchant, line 86: veriblock.MaxContextCountVbkPublication is bad name for this variable.
It means "maximum allowed VBK blocks per ALT block". Find a better name.

Same for max number of VTBs per ALT block, and max number of ATVs per ALT block

Warchant: I'd call it "MaxVbkBlocksPerAltBlock" or something like that

Warchant: Also double check their values.

PR #513, opened 2021-01-11
Feature/dbiterator

Mr-Leshiy:


include/veriblock/storage/block_iterator.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+  virtual void next() = 0;
+
+  virtual BlockIndex<BlockT> value() const = 0;
+
+  virtual hash_t key() const = 0;
+
+  virtual bool valid() const = 0;
+
+  virtual bool seek_start() = 0;
```
Warchant, line 42: please write docs for these funcs.
can seek_start fail? if not, why does it return `bool`?

include/veriblock/storage/block_provider.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * An abstraction over on-disk storage iterator.
+ *
+ * veriblock-pop-cpp does not dictate how to iterate over the storage. Altchain
+ * must create derived class.
+ *
+ * @ingroup interfaces
+ */
+
+template <typename BlockT>
+struct BlockProvider {
```
Warchant, line 32: I think this interface should be hidden in namespace `details` and then there would be another interface which implements this interface specific for each block, e.g.
```
struct BlockProvider : public GenericBlockProvider<AltBlock>, public GenericBlockProvider<VbkBlock>, public GenericBlockProvider<BtcBlock> {}
```

src/storage/util.cpp
```diff
@@ -2,11 +2,64 @@
...
+
+  std::vector<index_t> blocks;
+
+  for (; it->valid(); it->next()) {
+    blocks.push_back(it->value());
+  }
+
+  hash_t tip_hash;
+  if (!provider.getTipHash(tip_hash)) {
+    return state.Invalid("bad-provider", "cannot get tip");
```
Warchant, line 33:
```suggestion
    return state.Invalid("bad-tip", fmt::format("Can not read tip of {} block tip", block_t::name()));
```

src/storage/util.cpp
```diff
@@ -2,11 +2,64 @@
...
+  auto* t = out.getBestChain().tip();
+  VBK_ASSERT(t != nullptr);
+
+  return true;
+}
+
+bool LoadAllTrees(AltBlockTree& tree,
+                  BlockProvider<BtcBlock>& btc_provider,
+                  BlockProvider<VbkBlock>& vbk_provider,
+                  BlockProvider<AltBlock>& alt_provider,
```
Warchant, line 49: pass single provider here, then it will be just casted to specific types when passed by ref

Warchant (changes requested):

src/storage/util.cpp
Mr-Leshiy: fixed

src/storage/util.cpp
```diff
@@ -2,11 +2,65 @@
...
+  using index_t = typename BlockTreeT::index_t;
+  using hash_t = typename index_t::hash_t;
+
+  std::vector<index_t> blocks;
+
+  auto it = provider.getBlockIterator();
+  for (it->seek_start(); it->valid(); it->next()) {
+    index_t val;
+    if (!it->value(val)) {
+      return state.Invalid("bad-it", "Can not get block value");
```
Warchant, line 25:
```suggestion
      return state.Invalid("bad-value", "Can not get block value");
```

src/storage/util.cpp
```diff
@@ -2,11 +2,65 @@
...
+    }
+    blocks.push_back(val);
+  }
+
+  hash_t tip_hash;
+  if (!provider.getTipHash(tip_hash)) {
+    return state.Invalid("bad-tip",
```
Warchant, line 32:
```suggestion
    return state.Invalid(index_t::block_t::name() + "-bad-tip",
```

Warchant: do same for "bad-value"

Warchant (approved):

src/storage/util.cpp
Mr-Leshiy: fixed

src/storage/util.cpp
Mr-Leshiy: fixed

PR #528, opened 2021-01-20
Added test framework class

Skemba:


pypoptools/pypoptesting/test_framework.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+class PopIntegrationTestFramework(metaclass=PopIntegrationTestMetaClass):
+    """Base class for a pop integration test script.
+    Individual pop integration test scripts should subclass this class and override the
+    set_test_params(), run_test() and setup_network() methods.
+    Individual tests can also override the following methods to customize the test setup:
+    - add_options()
+    - setup_chain()
+    - setup_nodes()
```
Warchant, line 61: lets stay with `setup_nodes` only, as others are altchain-specific

pypoptools/pypoptesting/test_framework.py
```diff
@@ -0,0 +1,182 @@
...
+    This class also contains various public and private helper methods."""
+
+    def __init__(self):
+        """Sets test framework defaults. Do not override this method. Instead, override the set_test_params() method"""
+        self.success = False
+        self.nodes = []
+        self.num_nodes = 0
+        self.set_test_params()
+
+    def main(self, factory_lambda: Callable[[], Node]):
```
Warchant, line 72:
```suggestion
    def main(self, factory_lambda: Callable[[Path], Node]):
```
We should pass path to `/tmp/$dir/node0` in factory, so that altchain will be configured on top of that dir

pypoptools/pypoptesting/test_framework.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def stop_nodes(self):
+        """Stop multiple test nodes"""
+        for node in self.nodes:
+            # Issue RPC to stop nodes
+            node.stop()
+
+    def _create_nodes_(self, factory_lambda: Callable[[], Node]):
+        for i in range(self.num_nodes):
+            self.nodes.append(factory_lambda)
```
Warchant, line 159: here you make a list of factories, whereas you should call factory to get Node instance and then append it to this list

pypoptools/pypoptesting/test_framework.py
Warchant: Path is from https://docs.python.org/3/library/pathlib.html

pypoptools/pypoptesting/test_framework.py
Warchant: so should be 

```
self.nodes.append(factory_lambda(dir))
```

Warchant (approved):

PR #593, opened 2021-02-17
[BREAKING] Fix merkle tree

Warchant:


include/veriblock/entities/merkle_tree.hpp
```diff
@@ -153,8 +153,13 @@ struct BtcMerkleTree : public MerkleTree<BtcMerkleTree, uint256> {
   }
 
   hash_t finalizeRoot() {
-    VBK_ASSERT(!layers.empty());
-    return layers[0][0].reverse();
+    if (layers.empty()) {
```
nviktorov, line 156: Let's add the same check for VbkMerkleTree

Warchant: Added

PR #600, opened 2021-02-22
ALT-388: Refactor MockMiner

nviktorov:


pypoptools/pypoptesting/framework/pop_util.py
```diff
@@ -39,9 +39,10 @@ def endorse_block(node: Node, apm, height: int, address: str = None) -> str:
         node.submitpopvbk(vbk_block.toVbkEncodingHex())
     for vtb in payloads.vtbs:
         node.submitpopvtb(vtb.toVbkEncodingHex())
-    node.submitpopatv(payloads.atv.toVbkEncodingHex())
+    for atv in payloads.atvs:
+        node.submitpopatv(atv.toVbkEncodingHex())
 
-    return payloads.atv.getId()
+    return payloads.atvs[0].getId()
```
Warchant, line 45: what if we have many ATVs here?

test/commands/atomicity_test.cpp
```diff
@@ -153,15 +153,15 @@ TEST_F(AtomicityTestFixture, AddVTB) {
   // endorsed
   auto vbk5 = popminer->vbk().getBestChain().tip()->getAncestor(5);
   ASSERT_TRUE(vbk5);
-  auto vbkpoptx1 = popminer->createVbkPopTxEndorsingVbkBlock(
-      vbk5->getHeader(), getLastKnownBtcBlock());
-  auto vbkpoptx2 = popminer->createVbkPopTxEndorsingVbkBlock(
-      vbk5->getHeader(), getLastKnownBtcBlock());
+  auto vbkpoptx1 =
+      popminer->endorseVbkBlock(vbk5->getHeader(), getLastKnownBtcBlock());
```
Warchant, line 157: I think `endorseVbkBlock` is less descriptive name than `createVbkPopTxEndorsingVbkBlock`. I'd revert back to old name

test/e2e/pop_fr_invalid_vbk_chain_test.cpp
```diff
@@ -23,10 +23,10 @@ TEST_F(PopFrInvalidVbkChainTest, SendInvalidVTBtoAlternativeVBKchain) {
 
   // next block in B endorses block number 25 in its chain twice, and
   // containing is block 39
-  auto vbkpoptx1 = popminer->createVbkPopTxEndorsingVbkBlock(
-      tipB->getAncestor(25)->getHeader(), getLastKnownBtcBlock());
-  auto vbkpoptx2 = popminer->createVbkPopTxEndorsingVbkBlock(
-      tipB->getAncestor(25)->getHeader(), getLastKnownBtcBlock());
+  auto vbkpoptx1 = popminer->endorseVbkBlock(tipB->getAncestor(25)->getHeader(),
```
Warchant, line 26: same

pypoptools/pypopminer/entities.cpp
```diff
@@ -152,4 +152,10 @@ void init_entities() {
       .add_property(
           "difficulty", &VbkBlock::getDifficulty, &VbkBlock::setDifficulty)
       .add_property("nonce", &VbkBlock::getNonce, &VbkBlock::setNonce);
+
+  class_<PopData, boost::shared_ptr<PopData>>("PopData")
+      .def("__repr__", &PopData::toPrettyString)
+      .def_readonly("atvs", &PopData::atvs)
+      .def_readonly("vtbs", &PopData::vtbs)
+      .def_readonly("context", &PopData::context);
```
Warchant, line 160: add a method `serialize()` which will return a list of 3 lists:
```
[["0ab31..12"], ["0ab31..12"], ["0ab31..12"]]
```

Warchant: you can do that if you define a function
```
list Serialie(const PopData& self) {
 ...
}
```

then add Serialize here

```
.def("serialize", &Serialize)
```

Warchant (changes requested):

pypoptools/pypoptesting/framework/pop_util.py
nviktorov: Actually, `endorseAltBlock` always return exactly one ATV. 
Just because we now use `PopData` structure with collection `atvs`, not single field `atv`, we have such an issue.

test/commands/atomicity_test.cpp
nviktorov: Yeah, that's what I'm talking about. I don't know how to name function clearly, which:
- create BtcTx with published VbkBlock
- mine just created BtcTx
- create VbkPopTx with containing BtcTx and BtcBlock of proof.
I thought that `createVbkPopTxEndorsingVbkBlock` is also not clear enough, because it doesn't only create VbkPopTx, but perform a half of endorsement by mining in BTC as well. We also have atomic `createVbkPopTxEndorsingVbkBlock` function which only creates VbkPopTx without any side effects.
Maybe you have any alternative names? Or think `createVbkPopTxEndorsingVbkBlock` is OK for now?

nviktorov: Renamed for now.
If we'll come with more appropriate name, we'd change it.

pypoptools/pypopminer/entities.cpp
nviktorov: Added

Warchant (approved):

PR #605, opened 2021-02-23
Fix pop activation height in test

enfipy:


Warchant (changes requested):
pop activation height will be different for different chains, so it makes sense to remove assert completely

Warchant (approved):

PR #610, opened 2021-02-25
Improve mempool

Mr-Leshiy:


Warchant (changes requested):
There's memory error in mempool test...

Warchant (approved):

PR #644, opened 2021-03-12
Increase generate_endorsed() timeout

Mr-Leshiy:


pypoptools/pypoptesting/framework/generate_util.py
```diff
@@ -11,7 +11,7 @@
 def generate_endorsed(node: Node, mock_miner,
                       atvs: int, vtbs: int,
                       alt_blocks: int, vbk_blocks: int, btc_blocks: int,
-                      seed: int, timeout=60) -> None:
+                      seed: int, timeout=6000) -> None:
```
Warchant, line 14: don't set it to be too big, otherwise it would not be possible to distinguish between slow/failing code

src/pop/mempool.cpp
```diff
@@ -292,6 +296,8 @@ MemPool::SubmitResult MemPool::submit<VbkBlock>(
     const std::shared_ptr<VbkBlock>& blk, ValidationState& state) {
   VBK_ASSERT(blk);
 
+  VBK_LOG_DEBUG("blk: %s", HexStr(blk->toVbkEncoding()));
```
Warchant, line 299: pls delete

Warchant (approved):

PR #684, opened 2021-04-13
Add python node altchain adaptors

Mr-Leshiy:


pypoptools/pypoptesting/test_runner.py
```diff
@@ -10,7 +10,6 @@
 def create_node(number: int, path: pathlib.Path) -> Node:
     return VBitcoindNode(number=number, datadir=path)
 
-
```
nviktorov, line 13: PEP 8: E305 expected 2 blank lines after class or function definition, found 1

PR #701, opened 2021-04-22
add support for "known blocks" implemented as deleted block indexes; streamline block index creation and deletion

black0rwhite:
A known block is a block that is not part of the block tree, but is stored in the in-mem block store and can be (re-)added to the block tree by restore()'ing it and its ancestors.
Removing a block from the block tree converts it into a known block.

This should reduce the number of times block hashes are calculated and allow algorithms to operate on multiple chains at once and attach metadata to non-best chain blocks.


include/veriblock/pop/blockchain/base_block_tree.hpp
```diff
@@ -832,8 +897,6 @@ struct BaseBlockTree {
  protected:
   //! stores ALL blocks, including valid and invalid
   block_index_t blocks_;
-  //! stores all removed blocks, to ensure pointers to blocks remain stable
-  block_index_t removed_;
```
Warchant, line 836: is there any particular reason to do this?

black0rwhite: yes:
* less block hash recalculations. block hash is needed to move a block index to and from removed_. commands can resolve blocks early and operate on pointers instead. endorsements can point to blocks directly instead of storing hashes.
* tracking of blocks that are known to the altchain but don't belong to the best chain including VBK and BTC blocks, blocks in the mempool(temp block tree)

Warchant: okay, but why it is necessary to store these blocks in `blocks_`? 
Why not separate map?

black0rwhite: I'd rephrase that question as: whether a block deleted or not is just an one-bit flag, why store this bit of information in a separate map?
There are many disadvantages of having a separate map:

* deleting and restoring a block is more expensive as the hash has to be recalculated(or cached like what is currently done);
* searching for the corresponding blockindex across 2 maps is more expensive;
* 2 maps will consume more RAM for the same workload as it's expensive to shrink/grow hash maps, or alternatively tree-based maps will not provide amortized O(1) complexity;
* checking whether the block is deleted is expensive;
* data is best kept normalized, and that means it would be advantageous to go further and merge isDeleted flag and the reference counter such that isDeleted = (ref_counter == 0), as the alternative is more asserts, performance loss and potential for mistakes.

Mr-Leshiy:
@black0rwhite , thank you for the suggestion. But can you pls update this PR and left only features with the `BLOCK_DELETED` and `removed_` map but with the minimal changes in the API. So I mean do not change `touchBlockIndex` function name, renaming `size` function to the `getNonDeletedBlockCount`. 
So the main intention to make the minimal changes only related to this feature at this moment, without any other modifications.
As I see it is possible to add a few new functions, `isDeleted` as a getter for blockindex and `findBlockIndex` but only the const version. Without the following one.
```
  template <typename T,
            typename = typename std::enable_if<
                std::is_same<T, hash_t>::value ||
                std::is_same<T, prev_block_hash_t>::value>::type>
  index_t* findBlockIndex(const T& hash) {
    const auto& t = as_const(*this);
    return const_cast<index_t*>(t.findBlockIndex(hash));
  }
  ```

**Review by Warchant — APPROVED** [review]
overall, looks ok. 

I'd recommend testing this change with e2e+functional tests on vbtc

File: include/veriblock/pop/blockchain/base_block_tree.hpp [hunk 2]
```diff
@@ -138,49 +171,56 @@ struct BaseBlockTree {
...
     // we can not load a block, which already exists on chain and is not a
     // bootstrap block
-    if (current && !current->hasFlags(BLOCK_BOOTSTRAP)) {
-      return state.Invalid(
-          "block-exists",
-          "Found duplicate block, which is not bootstrap block");
-    }
+    if (current) {
+      if (current->isDeleted()) {
+        current->restore();
```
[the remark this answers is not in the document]
black0rwhite: Just like the constructor. Ideally, only the block trees should be able to call it.

Mr-Leshiy:
@black0rwhite , will merge this PR into the separate branch and test this version with the vBTC and vETC, after will create a new PR and assign you as reviewer

PR #715, opened 2021-04-29
move payload->command conversion code out of PayloadsStorage into tree-spefic CommandGroupStore classes; allow fetching a single command; make it easy to add caching and early payload validation

black0rwhite:


Mr-Leshiy:
@black0rwhite , will look into this PR after the https://github.com/VeriBlock/alt-integration-cpp/pull/701, because there are some identical modification exists. So will need update in future.

include/veriblock/pop/blockchain/blockchain_util.hpp
```diff
@@ -152,6 +152,21 @@ void payloadToCommands(Tree& tree,
...
+template <typename Tree, typename Payload>
+void payloadToCommandGroup(Tree& tree,
+                           const Payload& payload,
+                           const std::vector<uint8_t>& containingHash,
+                           CommandGroup& cg) {
+  // const auto& pl = tree.getPayloadsIndex();
+
+  cg.payload_type_name = &Payload::name();
+  cg.id = payload.getId().asVector();
+  // cg.valid = pl.getValidity(containingHash, cg.id);
```
Warchant, line 165: ?

include/veriblock/pop/blockchain/blockchain_util.hpp
```diff
@@ -152,6 +152,21 @@ void payloadToCommands(Tree& tree,
 
 struct PopData;
 
+//! @private
+template <typename Tree, typename Payload>
+void payloadToCommandGroup(Tree& tree,
+                           const Payload& payload,
+                           const std::vector<uint8_t>& containingHash,
+                           CommandGroup& cg) {
+  // const auto& pl = tree.getPayloadsIndex();
```
Warchant, line 161: ?

include/veriblock/pop/blockchain/blockchain_util.hpp
black0rwhite: the validity index is not functional ATM, so I disabled it to remove the associated overhead

Warchant: why not removed? why is it commented?

black0rwhite: it's potentially useful code + a reminder that we have CommandGroup::valid member

Mr-Leshiy:
@black0rwhite , can you describe more detailed what is the intention to move functionality from the PayloadsStorage into the two separate classes ? What benefits this modification will bring ? 
Also have questions with this lines of code:
```
// const auto& pl = tree.getPayloadsIndex();
  cg.payload_type_name = &amp;Payload::name();
   cg.id = payload.getId().asVector();
  // cg.valid = pl.getValidity(containingHash, cg.id);
```
Why we do not need to use `valid` field anymore ? Or maybe it will initialized later ? Because at this point it looks scary just not to initialize previously usable field. Can you provide some insights of this. Maybe I have missed something.

And also about this one `make it easy to add caching and early payload validation` can you provide details of it. As I understood this feature is not presented in this PR. Is it possible to implement it in the current version of the PayloadsStorage ?

black0rwhite:
> @black0rwhite , can you describe more detailed what is the intention to move functionality from the PayloadsStorage into the two separate classes ? What benefits this modification will bring ?

PayloadsStorage is supposed to be provided by the alchain. Its abstract `getATV/VTB/VBK` and `writePayloads` methods form a storage interface.

The pair of `getCommands()` methods perform an orthogonal task of converting payloads to command groups. This functionality is provided by the integration library which implements the methods which to my best knowledge are not overridden by altchains. One more clue that it doesn't belong here is the fact that commands and command groups are internal data structures/implementation detail that we don't want to expose to the altchain and vice versa. In fact, `Command` and `CommandGroup` are marked private in the API doc.

The benefits are:
  * not cluttering the public API with implementation details
  * not having to change the public API over and over again due to implementation changes
  * putting the library in charge of the command group creation, caching and/or other optimizations as it is in a much better position to know what to do and how to do it.

> 
> And also about this one `make it easy to add caching and early payload validation` can you provide details of it.

Early payload validation = running some(ideally all) command validation code in the command constructor instead of `Execute()`/`Unexecute()`. Together with caching and the `known blocks` feature this allows us to run expensive checks and block look-ups only once per block application+unapplication or even multiple such operations depending on the size of the cache. The size of the cache gets smaller due to blocks being referenced by pointers instead of hashes. Additionally, `early payload validation` and `known blocks` allow us to efficiently store endorsements as pointers to blocks instead of hashes, which in turn allows us to get rid of `allHashesInChain` in the fork resolution code and speed up certain worst-case scenarios.

> As I understood this feature is not presented in this PR. Is it possible to implement it in the current version of the PayloadsStorage ?

Yes, although that would still require a change to the signature and implementation of `getCommands()`, in fact multiple changes

black0rwhite:
> Also have questions with this lines of code:
> 
> ```
> // const auto& pl = tree.getPayloadsIndex();
>   cg.payload_type_name = &amp;Payload::name();
>    cg.id = payload.getId().asVector();
>   // cg.valid = pl.getValidity(containingHash, cg.id);
> ```
> 
> Why we do not need to use `valid` field anymore ? Or maybe it will initialized later ? Because at this point it looks scary just not to initialize previously usable field. Can you provide some insights of this. Maybe I have missed something.

Yes, the payload validity map was not implemented correctly and is not used. I did not remove this code because it would be fine if the payload map gets implemented correctly. I commented it out because querying the validity map that has no information nevertheless introduces an overhead. This query won't be optimized away by a compiler. Perhaps, I should have explained it in a comment. Reference:
https://github.com/VeriBlock/alt-integration-cpp/blob/d5b044c21d1b1254e901f7088fe1368647e61c05/include/veriblock/pop/storage/payloads_index.hpp#L25

Mr-Leshiy (changes requested):
@black0rwhite , got it, Generally looks great to me.
But can you also update and remove the unused functions getCommands from the PayloadStorage.

black0rwhite:
Changes besides a rebase:
* removed getCommands from PayloadStorage
* added a note to the commented out code
* merged another patch in that migrates the VBK tree from getCommands to getCommand as an optimization and enables pre-validation

Warchant (approved):
lgtm. wait for @Mr-Leshiy approve

Mr-Leshiy (approved):

PR #734, opened 2021-05-07
[Updated] add support for "known blocks" implemented as deleted block indexes; streamline block index creation and deletion

Mr-Leshiy:


Mr-Leshiy:
@black0rwhite , pls look at this PR.
I have updated getBlocks() function which will return just non-deleted blocks as before, so the meaning of this public functions will not change

PR #744, opened 2021-05-14
Add getMissingBtcBlocks() method into the mempool

Mr-Leshiy:


include/veriblock/pop/mempool.hpp
```diff
@@ -330,12 +332,7 @@ struct MemPool {
       auto& pl = **it;
       ValidationState state;
       auto valid = mempool_tree_.checkContextually(pl, state);
-      if (!valid) {
-        remove(pl);
-        it = c.erase(it);
-      } else {
-        ++it;
-      }
+      it = !valid ? (remove(pl), c.erase(it)) : std::next(it);
```
blackyblack, line 335: Why this change? It looks worse than before.

src/pop/blockchain/pop/vbk_block_tree.cpp
```diff
@@ -211,7 +213,10 @@ bool VbkBlockTree::validateBTCContext(const VbkBlockTree::payloads_t& vtb,
                                return height <= vtb.containingBlock.getHeight();
                              });
 
-  return isValid ? true : state.Invalid("block-referenced-too-early");
+  return isValid
+             ? (invalid_vtbs.erase(vtb.getId()), true)
+             : (invalid_vtbs[vtb.getId()].missing_btc_block = connectingHash,
+                state.Invalid("block-referenced-too-early"));
```
blackyblack, line 219: Normal `if` should look better here:
```
if (isValid) {
  invalid_vtbs.erase(vtb.getId());
  return true;
}
```

src/pop/blockchain/pop/vbk_block_tree.cpp
```diff
@@ -211,7 +213,10 @@ bool VbkBlockTree::validateBTCContext(const VbkBlockTree::payloads_t& vtb,
                                return height <= vtb.containingBlock.getHeight();
                              });
 
-  return isValid ? true : state.Invalid("block-referenced-too-early");
+  return isValid
+             ? (invalid_vtbs.erase(vtb.getId()), true)
```
blackyblack, line 217: Is it possible that missing BTC block will be provided with another VTB? Those bad VTBs will stay forever in such a case.
You can try to store missing BTC blocks in a flat structure, eg std::vector and check if provided BTC context fixes the gap.

PR #749, opened 2021-05-18
VBK finalization test

blackyblack:


include/veriblock/pop/blockchain/base_block_tree.hpp
```diff
@@ -690,13 +690,11 @@ struct BaseBlockTree {
   //!
   //! @returns false if block not found or prereq are not met
   //! @private
-  virtual bool finalizeBlockImpl(const hash_t& hash,
+  virtual bool finalizeBlockImpl(index_t* index,
```
Mr-Leshiy, line 693: Maybe should change it from ptr to ref, so we can remove the validation that ptr does not initialized

blackyblack: I had the same idea but it requires additional dereference in tests and some pointer comparisons will have to be redone.

Mr-Leshiy (approved):

PR #753, opened 2021-05-20
Fix Max PopTxs vulnerability

Mr-Leshiy:


src/pop/blockchain/pop/vbk_block_tree.cpp
```diff
@@ -234,6 +234,13 @@ bool VbkBlockTree::addPayloadToAppliedBlock(index_t& index,
     return true;
   }
 
+  if (index.getPayloadIds<payloads_t>().size() + 1 >
+      MAX_VBKPOPTX_PER_VBK_BLOCK) {
```
blackyblack, line 238: `if (index.getPayloadIds<payloads_t>().size() >= MAX_VBKPOPTX_PER_VBK_BLOCK)`

blackyblack (approved):

src/pop/blockchain/pop/vbk_block_tree.cpp
Mr-Leshiy: fixed

PR #771, opened 2021-05-27
untangle the continueOnInvalid hack; reimplement filterInvalidPayloads based on an incremental block builder; loadBlock: remove stateful duplicate enforcement in unconnected blocks

black0rwhite:


Warchant (approved):

test/pop/storage/save_load_tree_test.cpp
```diff
@@ -57,6 +57,52 @@ TEST_F(SaveLoadTreeTest, ValidTree) {
   assertTreesEqual();
 }
 
+TEST_F(SaveLoadTreeTest, ReloadWithoutDuplicates_NoUnconnectedBlockCheck) {
```
Mr-Leshiy, line 60: Can you describe this test case more clear. What situation we are trying to check ?

src/pop/blockchain/alt_block_tree.cpp
```diff
@@ -409,10 +531,20 @@ bool AltBlockTree::loadBlock(const stored_index_t& index,
...
   if (!checkIdsForDuplicates<ATV>(atvIds, state)) return false;
 
-  if (hasDuplicates<VbkBlock>(*current, vbkblockIds, *this, state) ||
-      hasDuplicates<VTB>(*current, vtbIds, *this, state) ||
-      hasDuplicates<ATV>(*current, atvIds, *this, state)) {
-    return false;
+  if (current->pprev != nullptr) {
+    // if the block is not yet connected, defer the stateful duplicate check to
+    // connectBlock()
+    if (current->isConnected()) {
```
Mr-Leshiy, line 537: As we have the same validation inside the connectBlock() why we can not remove validation from this place ?

Mr-Leshiy:
Am I right that you just move the current popdata filtering process from the altblocktree directly to the another entity  BlockPayloadMutator ?
And the question is how it is working with the pop_state_machine. So previously we do not separate this validation from the tree because for the full validation of the pop data we need to emulate the entire validation process including execution of the pop state machine checks. How is it standing with the BlockPayloadMutator ?

black0rwhite:
> Am I right that you just move the current popdata filtering process from the altblocktree directly to the another entity BlockPayloadMutator ?
The filtering logic used to be spread over the mempool tree, altblocktree, fork resolution comparator and the state machine.
This patch puts all the logic into the mempool tree.

`BlockPayloadMutator` is not a filtering entity, more like a leaf block constructor and validator -- that's what its current use cases are. It operates on the edge between unconnected blocks(which are freely modifiable, aren't significantly restricted and  checked extensively) and non-leaf connected blocks which are immutable.

> And the question is how it is working with the pop_state_machine. So previously we do not separate this validation from the tree because for the full validation of the pop data we need to emulate the entire validation process including execution of the pop state machine checks. How is it standing with the BlockPayloadMutator ?

Its interaction with `PopStateMachine` is similar to `VbkBlockTree::AddPayloads`. It can only be used to add payloads if they don't invalidate any block or payload. The end effect of adding a valid block via `acceptBlock` and activating it via `setState` is identical to adding an empty or partially populated block, activating it via `setState` and adding (the rest of) its payloads using `BlockPayloadMutator`.

I has to be an object, and not a function because it needs to cache validation-related data for efficiency.

src/pop/blockchain/alt_block_tree.cpp
black0rwhite: As far as I understand this is just a sanity check that is supposed to verify that connectBlock did its job. I think I see some other issues with loadBlock, but that belongs to another patch. I fixed this specific issue since it was surfaced by this patch.

test/pop/storage/save_load_tree_test.cpp
Mr-Leshiy: Some am I right that second block which will contain duplicated payloads actually marked as invalid in the tree and this block can be stored in the tree even marked as invalid and such test checks that load() function will not fail and correctly load such invalid block ?

PR #783, opened 2021-06-04
Feature/split load block call alt 580

blackyblack:


black0rwhite:
Perhaps, loadLeafBlock and loadRootBlock would be more descriptive than loadBlockForward and loadBlockBackward?

While I don't know the intent and context of these changes, I think that loading leaf blocks before the root one is impractical considering the upcoming change towards storing block references as pointers instead of hashes in endorsements. Such a change disallows dangling references and thus loading blocks in the leaves-to-root order will have to happen in 2 passes:
1) instantiate empty blocks
2) populate endorsements and possibly other data

Mr-Leshiy:
@black0rwhite  The intention of this PR is to split the current functionality of the loadBlock and narrowing the scope of the each method.

include/veriblock/pop/blockchain/base_block_tree.hpp
```diff
@@ -157,89 +157,7 @@ struct BaseBlockTree {
    * state.
    */
   virtual bool loadBlock(const stored_index_t& index, ValidationState& state) {
```
Mr-Leshiy, line 159: Remove this function, just leave only loadBlockForward loadBlockBackward and make them `public` and `virtual` as with the loadBlock

include/veriblock/pop/blockchain/blocktree.hpp
```diff
@@ -161,6 +162,35 @@ struct BlockTree : public BaseBlockTree<Block> {
...
+
+    VBK_ASSERT(current->pnext.size() == 1);
+    auto* next = *current->pnext.begin();
+
+    // recover chainwork
+    auto chainWork = getBlockProof(next->getHeader());
+    current->chainWork = next->chainWork - chainWork;
+
+    // clear blockOfProofEndorsements inmem field
+    current->clearBlockOfProofEndorsement();
```
blackyblack, line 188: I think it is better to avoid duplicate code by using loadBlockInner.

PR #789, opened 2021-06-09
create a single PopStateMachine instance per PopAwareForkResolutionComparator; pass the protected tree via the constructor as the protected and protecting trees are coupled anyway; minor cleanups

black0rwhite:


include/veriblock/pop/blockchain/pop/fork_resolution.hpp
```diff
@@ -444,33 +444,35 @@ struct PopAwareForkResolutionComparator {
                                        protecting_params_t,
                                        protected_params_t>;
 
-  PopAwareForkResolutionComparator(std::shared_ptr<ProtectingBlockTree> tree,
+  PopAwareForkResolutionComparator(ProtectedBlockTree& ed,
```
Mr-Leshiy, line 447: What was the intention to move ProtectedBlockTree& ed as an argument of the functions to the field of this struct ? What benefits these should bring to us ?

black0rwhite: This is just a clean up, so the benefits are small. Basically, the states of the protected and protecting trees are coupled. Any pop state machine call on the protected tree changes the state of the protecting tree. So there's no way we could have 2 protected trees per 1 protecting tree. This change leads to a minor simplification of code.

Mr-Leshiy:
LGTM, after update to the latest develop we can merge it

Mr-Leshiy (approved):

PR #812, opened 2021-07-06
Update Blob struct constructors 

Mr-Leshiy:


include/veriblock/pop/blob.hpp
```diff
@@ -221,22 +220,21 @@ struct Blob {
...
 
  protected:
-  inline void assign(Slice<const uint8_t> slice) {
-    if (slice.size() > N) {
-      throw std::invalid_argument(
-          "Blob(): invalid data size: " + std::to_string(slice.size()) + " > " +
-          std::to_string(N));
-    }
-    std::copy(slice.begin(), slice.end(), data_.begin());
+  inline void assing(const std::initializer_list<uint8_t>& list) {
```
blackyblack, line 223: assign

blackyblack (changes requested):

include/veriblock/pop/blob.hpp
Mr-Leshiy: fixed

blackyblack (approved):

PR #819, opened 2021-07-13
Prepare for finalization test

blackyblack:


pypoptools/pypoptesting/tests/pop/finalize.py
```diff
@@ -0,0 +1,67 @@
...
+        expected_best = self.nodes[0].getbestblock()
+        best_blocks = [node.getbestblock() for node in self.nodes]
+        for best in best_blocks:
+            assert best == expected_best
+
+        self.log.info("nodes successfully synced")
+        self.log.info("trying getblock {}".format(last_block + 1))
+
+        erased_block_hash = self.nodes[0].getblockhash(last_block + 1)
+        erased_block = self.nodes[0].rpc.getblock(erased_block_hash)
```
Mr-Leshiy, line 65: We can not cal an rpc function directly, because in the ETC we have totally different rpc function`s names so this will fail.
Need to call `self.nodes[0].getblock(erased_block_hash)

Mr-Leshiy (changes requested):

pypoptools/pypoptesting/tests/pop/finalize.py
blackyblack: Fixed

pypoptools/pypoptesting/tests/pop/finalize.py
```diff
@@ -0,0 +1,79 @@
...
+
+        expected_best = self.nodes[0].getbestblock()
+        best_blocks = [node.getbestblock() for node in self.nodes]
+        for best in best_blocks:
+            assert best == expected_best
+
+        self.log.info("nodes successfully synced")
+        self.log.info("trying getblock {}".format(last_block + 1))
+
+        erased_block_hash = self.nodes[0].getblockhash(last_block + 1)
```
Mr-Leshiy, line 76: We can reuse the known `before_erased_block_hash`

Mr-Leshiy (changes requested):

pypoptools/pypoptesting/tests/pop/finalize.py
blackyblack: Fixed

Mr-Leshiy (approved):

PR #837, opened 2021-08-30
fix 4_chains_converge_random function test case for any keystoneInterval

memekas:


Warchant:
Please create new branch with these changes from last develop branch.

PR #841, opened 2021-09-02
OverrideTip refactoring

blackyblack:


include/veriblock/pop/blockchain/pop/fork_resolution.hpp
```diff
@@ -552,6 +552,7 @@ struct PopAwareForkResolutionComparator {
       }
 
       VBK_LOG_DEBUG("Candidate contains VALID commands, chain B wins");
+      ed_.overrideTip(candidate);
```
Warchant, line 555: Why did you make this change?

include/veriblock/pop/blockchain/pop/fork_resolution.hpp
```diff
@@ -662,6 +663,7 @@ struct PopAwareForkResolutionComparator {
       }
 
       VBK_LOG_DEBUG("Chain B wins");
+      ed_.overrideTip(candidate);
```
Warchant, line 666: Also here

test/pop/blockchain/CMakeLists.txt
```diff
@@ -68,4 +68,3 @@ addtest(temp_block_tree_test temp_block_tree_test.cpp)
 addtest(alt_tree_bootstrap_test alt_tree_bootstrap_test.cpp)
 addtest(altblock_finalization_test altblock_finalization_test.cpp)
 addtest(vbkblock_finalization_test vbkblock_finalization_test.cpp)
-addtest(altblock_restoration_test altblock_restoration_test.cpp)
```
Warchant, line 71: why you removed this test?

include/veriblock/pop/storage/util.hpp
```diff
@@ -113,7 +113,7 @@ void saveTree(
   std::vector<const index_t*> dirty_indices;
 
   // map pair<hash, shared_ptr<index_t>> to vector<index_t*>
-  for (auto& index : tree.getAllBlocks()) {
+  for (auto& index : tree.getBlocks(false)) {
```
Warchant, line 116: I think it would be better to have a enum class 
```
enum class GetBlocksPolicy {
  DEFAULT,
  SKIP_DELETED
}

```

and set getBlocks(GetBlocksPolicy policy = GetBlocksPolicy::DEFAULT)

include/veriblock/pop/blockchain/pop/fork_resolution.hpp
blackyblack: Because I want activateBestChain to be consistent. It should always switch to the winner.

test/pop/blockchain/CMakeLists.txt
blackyblack: We never use block restoration code.

include/veriblock/pop/storage/util.hpp
blackyblack: A bit too verbose but I can do it this way.

blackyblack: Done

include/veriblock/pop/blockchain/alt_block_tree.hpp
```diff
@@ -115,19 +114,16 @@ extern template struct BaseBlockTree<AltBlock>;
...
+//! @note after AltBlockTree::activateBestChain AltBlockTree always corresponds to a state, as if winner chain have been applied.
 //!
 //!```cpp
 //!         if(result < 0) {
 //!             // candidate has better POP score.
 //!             // tree already switched to candidate chain.
 //!
-//!             UpdateTip(candidate->getHash());
-//!             // NOTE: update `tip`, otherwise old tip will be passed to first arg,
-//!             // and comparePopScore will die on assert
```
Warchant, line 130: Why did you delete this?
Users must update tip of their block tree, otherwise ALT tree in alt-cpp and block tree in altchain won't be in sync

src/pop/blockchain/alt_block_tree.cpp
```diff
@@ -546,8 +544,7 @@ void AltBlockTree::overrideTip(index_t& to) {
                  "the active chain tip(%s) must be fully valid",
                  to.toPrettyString());
 
-  onBeforeOverrideTip.emit(to);
```
Warchant, line 549: why did you remove this signal?

src/pop/blockchain/pop/vbk_block_tree.cpp
```diff
@@ -35,28 +35,30 @@ void VbkBlockTree::determineBestChain(index_t& candidate,
...
-    this->overrideTip(candidate);
+    VBK_LOG_DEBUG("Candidate chain won");
   } else {
     // the current chain is better
     VBK_LOG_DEBUG("Active chain won");
   }
 }
 
+bool VbkBlockTree::setState(const hash_t& block, ValidationState& state) {
+  return base::setState(block, state);
```
Warchant, line 54: I think you meant to do this

```
auto* index = getBlockIndex(hash);
if (!index) return false;
return this->setState(*index, state);
```

Warchant (changes requested):

src/pop/blockchain/pop/vbk_block_tree.cpp
Warchant: Please add a testcase to reveal this problem

include/veriblock/pop/blockchain/alt_block_tree.hpp
blackyblack: I thought it is related to overrideTip call which is obsolete after this refactoring.

src/pop/blockchain/alt_block_tree.cpp
blackyblack: It is not used anymore.

src/pop/blockchain/pop/vbk_block_tree.cpp
blackyblack: It will call abstract base setState which looks for blockIndex by hash and redirects to specialized version of setState. I cannot make a test because there is no such issue.

Warchant: anyway, it is not clear from the code. please update

Warchant:
Please rebase

Warchant:
Copied everything except rename in #926

PR #852, opened 2021-09-08
Update PopParamsTest python test

Mr-Leshiy:


Warchant (changes requested):
Why did you do this change?

PR #854, opened 2021-09-14
Test for ATV prioritization

blackyblack:


include/veriblock/pop/txfees.hpp
```diff
@@ -0,0 +1,18 @@
...
+
+#ifndef ALT_INTEGRATION_VERIBLOCK_TXFEES_HPP
+#define ALT_INTEGRATION_VERIBLOCK_TXFEES_HPP
+
+#include "entities/coin.hpp"
+#include "entities/vbktx.hpp"
+
+namespace altintegration {
+
+Coin calculateTxFee(const VbkTx& tx);
```
Warchant, line 14: I think this can be VbkTx method

src/pop/entities/vbktx.cpp
```diff
@@ -62,6 +62,16 @@ uint256 VbkTx::getHash() const {
   return sha256(stream.data());
 }
 
+Coin VbkTx::calculateTxFee() const {
+  decltype(Coin::units) output{0};
+  for (const auto& o : outputs) {
+    output += o.coin.units;
+  }
+  VBK_ASSERT_MSG(sourceAmount.units >= output,
```
Warchant, line 70: I think this should be removed, because otherwise malitious actor could send bad tx and all nodes will fail on this assert

Warchant: alternatively you can add validation of fees

Warchant: hmmmm. Okay. I think we need to do fee validation (check veriblock rules) and leave this assert here then

include/veriblock/pop/mempool.hpp
```diff
@@ -378,6 +378,23 @@ struct MemPool {
...
+  struct EndorsedAltComparator {
+    EndorsedAltComparator(const MemPool& parent) : parent_(parent) {}
+    int operator()(const std::shared_ptr<ATV>& a,
+                   const std::shared_ptr<ATV>& b) const;
+
+   private:
+    const MemPool& parent_;
+  };
+
+  struct TxFeeComparator {
```
Warchant, line 391: I thought we could have 1 comparator, which does comparison by TX FEE first, then if it is equal will continue with other criterias. How does multiple comparators align with this idea?

blackyblack: I prefer to do it in a functional way and construct combined comparator from separate comparators. Unfortunately C++ does not have something like std::combine, so no advantage here but still looks a bit better to me.

src/pop/mempool.cpp
```diff
@@ -36,6 +82,7 @@ PopData generatePopDataImpl(
 
     // try to fit ATVs
     auto& atvcandidates = block.second->atvs;
```
Warchant, line 84: hmm. I wonder if we can maintain max heap or something like that in `block.second->atvs`. Could you please explore?

blackyblack: Could you elaborate? Do you want to collect ATVs from different VBK blocks?

Warchant: `block.second` contains a vector of ATVs. I wonder if we could maintain `atvs` as sorted array instead of sorting it every time.

blackyblack: Interesting idea. I will take a look what data structure fits best for our case.

Warchant: I am pretty sure that "max heap" will work 
https://en.wikipedia.org/wiki/Min-max_heap

Mr-Leshiy (approved):

Warchant (approved):

PR #863, opened 2021-09-20
fork resolution: address potential null pointer dereference warnings

black0rwhite:


Warchant:
GCC still does not build

blackyblack:
> GCC still does not build

I think CI is fixed in my PR. So please put this on a hold.

PR #874, opened 2021-10-06
Change comparePopScore log level

Mr-Leshiy:
I assume we should change the log level for the comparePopScore result to the `VBK_LOG_DEBUG` or `VBK_LOG_INFO`


Warchant (changes requested):
I intentionally made it "WARN", please do not change

PR #885, opened 2021-11-06
Fix statefull duplication validation

Mr-Leshiy:
Fix for this issue https://github.com/VeriBlock/vbk-ri-btc/issues/593

### DIFF
--- a/include/veriblock/pop/blockchain/alt_chain_params.hpp
+++ b/include/veriblock/pop/blockchain/alt_chain_params.hpp
@@ -255,6 +255,10 @@ struct AltChainParams {
                                 const std::vector<uint8_t>& root,
                                 ValidationState& state) const noexcept = 0;
 
+  virtual bool isAncestor(
+      const AltBlock::hash_t& descendant_hash,
+      const AltBlock::hash_t& ancestor_hash) const noexcept = 0;
+
  public:
   std::shared_ptr<PopPayoutsParams> mPopPayoutsParams =
       std::make_shared<PopPayoutsParams>();
--- a/src/pop/blockchain/alt_block_tree.cpp
+++ b/src/pop/blockchain/alt_block_tree.cpp
@@ -443,11 +443,7 @@ void appendIds(std::unordered_set<std::vector<uint8_t>>& ids,
 
 AltBlockTree::BlockPayloadMutator::BlockPayloadMutator(tree_t& tree,
                                                        block_index_t& block)
-    : tree_(tree),
-      block_(block),
-      payload_index_(tree.getPayloadsIndex()),
-      // don't look for duplicates in the block itself
-      chain_(tree.getParams().getBootstrapBlock().height, block_.pprev) {
+    : tree_(tree), block_(block), payload_index_(tree.getPayloadsIndex()) {
   VBK_ASSERT_MSG(block_.pprev,
                  "Adding payloads to a bootstrap block is not allowed");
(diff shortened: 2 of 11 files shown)


src/pop/blockchain/alt_block_tree.cpp
```diff
@@ -468,10 +468,28 @@ AltBlockTree::BlockPayloadMutator::BlockPayloadMutator(tree_t& tree,
...
-  // make sure existing blocks do not contain this id
-  for (const auto& hash : payload_index_.getContainingAltBlocks(payload_id)) {
+template <>
+bool AltBlockTree::BlockPayloadMutator::isStatefulDuplicate<ATV>(
+    const ATV::id_t& atv_id) {
+  auto id = atv_id.asVector();
+  // make sure existing blocks from the active chain do not contain this id
+  for (const auto& hash : payload_index_.getContainingAltBlocks(id)) {
+    const auto* candidate = tree_.getBlockIndex(hash);
+    if (chain_.contains(candidate)) {
```
Warchant, line 478: For ATV we could see if "block of proof" is in VBK tree (not necessarily in active chain), if yes -> it will contain a list of IDs. If ID exists -> it's a duplicate.

src/pop/blockchain/alt_block_tree.cpp
```diff
@@ -468,10 +468,28 @@ AltBlockTree::BlockPayloadMutator::BlockPayloadMutator(tree_t& tree,
...
+}
+
+template <>
+bool AltBlockTree::BlockPayloadMutator::isStatefulDuplicate<VTB>(
+    const VTB::id_t& vtb_id) {
+  auto id = vtb_id.asVector();
+  // make sure existing blocks from the active chain do not contain this id
+  for (const auto& hash : payload_index_.getContainingAltBlocks(id)) {
     const auto* candidate = tree_.getBlockIndex(hash);
     if (chain_.contains(candidate)) {
```
Warchant, line 494: For VTB: find "containing block" and search for ID. If id exists -> it's duplicate.

Warchant: and we need to add new validation - check if these blocks (VTB containing, block of proof, ATV block of proof) are behind final - if they are, then they should fail validation

src/pop/blockchain/alt_block_tree.cpp
```diff
@@ -482,13 +500,21 @@ bool AltBlockTree::BlockPayloadMutator::isStatefulDuplicate(
   return false;
 }
 
+template <>
+bool AltBlockTree::BlockPayloadMutator::isStatefulDuplicate<VbkBlock>(
+    const VbkBlock::id_t& vbk_id) {
+  // make sure that provided VbkBlock does not contains in the existing
+  // VbkBlockTree state
+  return this->tree_.vbk().getBlockIndex(vbk_id) != nullptr;
```
black0rwhite, line 508: When we are performing stateful duplicate checks for alt block A payloads, the VBK tree view reflects block B, which isn't necessarily a direct ancestor, or an ancestor at all.

include/veriblock/pop/blockchain/alt_chain_params.hpp
```diff
@@ -255,6 +255,10 @@ struct AltChainParams {
                                 const std::vector<uint8_t>& root,
                                 ValidationState& state) const noexcept = 0;
 
+  virtual bool isAncestor(
```
Warchant, line 258: why this is not in BaseBlockTree?

PR #889, opened 2021-11-16
Add test for stateful duplicate issue

Warchant:


include/veriblock/pop/blockchain/base_block_tree.hpp
```diff
@@ -849,21 +850,22 @@ struct BaseBlockTree {
...
                   preserveBlocksBehindFinal);
 
     index_t* finalizedBlock = &index;
 
-    // prereq is not met - finalized block must be on active chain
-    if (!activeChain_.contains(finalizedBlock)) {
-      return state.Invalid("block-not-on-active-chain");
+    if (finalizedBlock == getRoot()) {
+      return true;
     }
```
blackyblack, line 863: Why this check?

src/pop/blockchain/alt_block_tree.cpp
```diff
@@ -568,6 +571,27 @@ bool AltBlockTree::setState(index_t& to, ValidationState& state) {
   return success;
 }
 
+void AltBlockTree::doFinalize() {
+  ValidationState state;
+  auto* tip = getBestChain().tip();
+  VBK_ASSERT(tip && "must be bootstrapped");
```
blackyblack, line 577: It seems you prefer to assert != nullptr. Make it consistent over new code:
```
VBK_ASSERT(tip != nullptr && "must be bootstrapped");
```

src/pop/blockchain/alt_block_tree.cpp
```diff
@@ -568,6 +571,27 @@ bool AltBlockTree::setState(index_t& to, ValidationState& state) {
...
+
+  int32_t maxReorg = (int32_t)getParams().getMaxReorgDistance();
+  if (maxReorg > getRoot().getHeight()) {
+    return;
+  }
```
blackyblack, line 582: It looks like finalization won't work until root reaches maxReorg height, eg if maxReorg is 100000 finalization won't work until root is on height 100000.

src/pop/blockchain/alt_block_tree.cpp
```diff
@@ -698,6 +724,8 @@ bool AltBlockTree::loadTip(const AltBlockTree::hash_t& hash,
     tip = tip->pprev;
   }
 
+  // finalize a block
+
```
blackyblack, line 728: What is this?

src/pop/blockchain/alt_block_tree.cpp
```diff
@@ -745,9 +773,11 @@ bool AltBlockTree::finalizeBlockImpl(index_t& index,
...
   auto* finalizedIndex = vbk().getBestChain()[firstBlockHeight];
   VBK_ASSERT_MSG(finalizedIndex != nullptr, "Invalid VBK tree state");
-  if (!vbk().finalizeBlock(*finalizedIndex, state)) {
-    return state.Invalid("vbktree-finalize-error");
-  }
+  bool result = vbk().finalizeBlock(*finalizedIndex, state);
+  VBK_ASSERT_MSG(result,
+                 "Failed to finalize block %s: %s",
+                 finalizedIndex->toShortPrettyString(),
+                 state.toString());
```
blackyblack, line 780: It seems that finalizeBlock should never fail. Put this assert inside finalizeBlock() and do not return anything from it.

blackyblack (changes requested):

include/veriblock/pop/blockchain/base_block_tree.hpp
Warchant: It's an early return

blackyblack: Yes but it is covered but the code below. Does not look like it makes any speed up.

src/pop/blockchain/alt_block_tree.cpp
Warchant: you're right, it's a typo. we should compare maxReorg with tip height, not root height

include/veriblock/pop/blockchain/base_block_tree.hpp
Warchant: this case will be triggered often, and instead of iterating over `tips` and doing some other looping we could just early return, this is faster by definition (we don't invoke code we don't need)

include/veriblock/pop/blockchain/base_block_tree.hpp
```diff
@@ -951,7 +952,7 @@ struct BaseBlockTree {
       ptr = ptr->pprev;
     }
 
-    return true;
+    return;
```
blackyblack, line 955: Useless.

include/veriblock/pop/blockchain/blocktree.hpp
```diff
@@ -211,15 +211,14 @@ struct BlockTree : public BaseBlockTree<Block> {
...
+  void finalizeBlock(index_t& index) {
     (void)index;
-    (void)state;
     // TODO: currently can not procced finalization for the vbk and btc trees,
     // need more investigation
 
     // return this->finalizeBlockImpl(
     //     index, param_->preserveBlocksBehindFinal(), state);
-    return true;
+    return;
```
blackyblack, line 221: Remove.

src/pop/blockchain/pop/vbk_block_tree.cpp
```diff
@@ -434,10 +433,8 @@ bool VbkBlockTree::finalizeBlockImpl(index_t& index,
   firstBlockHeight = std::max(bootstrapBlockHeight, firstBlockHeight);
   auto* finalizedIndex = btc().getBestChain()[firstBlockHeight];
   VBK_ASSERT_MSG(finalizedIndex != nullptr, "Invalid BTC tree state");
-  if (!btc().finalizeBlock(*finalizedIndex, state)) {
-    return state.Invalid("btctree-finalize-error");
-  }
-  return base::finalizeBlockImpl(index, preserveBlocksBehindFinal, state);
+  btc().finalizeBlock(*finalizedIndex);
+  return base::finalizeBlockImpl(index, preserveBlocksBehindFinal);
```
blackyblack, line 437: Remove `return`:
```
btc().finalizeBlock(*finalizedIndex);
base::finalizeBlockImpl(index, preserveBlocksBehindFinal);
```

test/pop/blockchain/altblock_finalization_test_3.cpp
```diff
@@ -0,0 +1,69 @@
...
+    altparam.mEndorsementSettlementInterval = 50;
+    altparam.mPreserveBlocksBehindFinal = 50;
+  }
+
+  std::vector<AltBlock> chain = {altparam.getBootstrapBlock()};
+};
+
+TEST_F(StatefulDuplicateWhenFinalized, VBK) {
+  alttree.onInvalidBlockConnected.connect(
+      [](BlockIndex<AltBlock> &/*index*/, ValidationState &state) {
```
blackyblack, line 28: Rename `state` - it hides PopTestFixture::state field.

blackyblack (changes requested):

Issue #890, opened 2021-11-17
Test for pop mempool inclusion order


Warchant:
Mempool returns PopData in this order:
1. Payloads with earlier VBK block will be earlier. VBK = use itself, ATV=block of proof, VTB=containing.
2. Payloads with VBK block on active chain will be earlier.
3. ATVs with same VBK block will be compared by ATV VBK tx fee. Higher fee - more chances that ATV will be picked.

Write a test that will check current logic against this ^.

PR #891, opened 2021-11-18
Test for pop mempool inclusion order

Mr-Leshiy:
issue https://github.com/VeriBlock/alt-integration-cpp/issues/890


Warchant:
Add cases when number of payloads is greater than a block limit.

Mr-Leshiy:
Such test case "ATVs with same VBK block will be compared by ATV VBK tx fee. Higher fee - more chances that ATV will be picked." already exists `getPop_txfeePriority`

Warchant (approved):

PR #896, opened 2021-11-25
Fix new mempool context gap issue

Mr-Leshiy:


Warchant (changes requested):
Wrong test.

Test should try to add VTB through mempool, whose containing block is X, where X < VBKTIP-12k (too old blocks window)

Warchant (approved):

PR #898, opened 2021-11-29
clean up BaseBlockTree::eraseBlock()

black0rwhite:


include/veriblock/pop/blockchain/base_block_tree.hpp
```diff
@@ -519,6 +519,15 @@ struct BaseBlockTree {
     tips_.insert(index);
   }
 
+  //! dangerous! permanently drops the block from the tree
+  void eraseBlockUnchecked(index_t& block) {
```
Warchant, line 523: Maybe rename to deallocateBlockUnsafe?

blackyblack: Why make a separate function for erasing block? You can merge `eraseBlock` and `eraseBlockUnchecked` into a single function.

black0rwhite: neither of eraseBlock functions is particularly safe. erase sounds unambiguous enough to me. drop might be also work. deallocate is a bit too low-level/implementation detail.

include/veriblock/pop/blockchain/base_block_tree.hpp
```diff
@@ -519,6 +519,15 @@ struct BaseBlockTree {
...
+    VBK_LOG_INFO("Deallocating block %s", block.toShortPrettyString());
+    auto count = blocks_.erase(makePrevHash(block.getHash()));
+    VBK_ASSERT_MSG(count == 1,
+                   "state corruption: block %s is not in the store",
+                   block.toPrettyString());
+  }
+
   //! dangerous! permanently drops the block from the tree
   void eraseBlock(index_t& block) {
```
Warchant, line 532: I am not convinced that Unsafe func is needed. It doesn't check that pnext is empty, which means that we can deallocate a block in a middle of a chain. 
I'd remove Unsafe version.

Warchant (changes requested):

include/veriblock/pop/blockchain/base_block_tree.hpp
black0rwhite: This is a refactoring commit. so, all it is intended to do is deduplicate code and regularize logging. 
Whether this function has a future depends on what happens to eraseBlock. We either want eraseLeaf+eraseRoot combo or a generalized eraseBlock that ensures we can only delete a root or a leaf. eraseLeaf/Root will have a use for eraseBlockUnchecked and I personally lean towards this approach as it is more explicit

include/veriblock/pop/blockchain/base_block_tree.hpp
black0rwhite: one more point regarding "deallocate": what would you call this function in java?

black0rwhite: > Why make a separate function for erasing block? You can merge eraseBlock and eraseBlockUnchecked into a single function.
The point of multiple erase functions is contracts/invariant enforcement via asserts.

blackyblack: > 

Well, eraseBlock should be eraseLeafBlock then. And eraseBlockUnchecked should be eraseAnyBlock.

black0rwhite: eraseBlock can erase the root block if the tree contains a single block, it's a bit of a mess atm. eraseBlockUnchecked was intended to sound unsafe as compared to eraseBlock.

blackyblack: If the tree contains a single block then root block is also a leaf block.

black0rwhite: hmm, although we don't explicitly check that the block is not the root, we assert that block.isDeleted(), so in a tree that isn't malformed, we can't use the current implementation of eraseBlock() to erase the root block.

black0rwhite: anyway, this commit wasn't intended to make erase* stuff perfect: https://github.com/VeriBlock/alt-integration-cpp/pull/898#discussion_r759164702

black0rwhite: As to whether the root block is a leaf block, yes, formally it is. However, deleting the root block is not something the developer expects when manipulating leaf blocks. eraseRoot/eraseLeaf split is formally eraseRoot/eraseLeafThatIsNotRoot, but that's a bit too noisy.

Warchant:
I am not convinced that this change is necessary. I am open for discussion if you think otherwise.

PR #899, opened 2021-12-02
Fix Logger undefined behaviour

Mr-Leshiy:


src/pop/logger.cpp
```diff
@@ -8,18 +8,16 @@
 
 namespace altintegration {
 
-static std::unique_ptr<Logger>& getLoggerInner() {
+static Logger* getLoggerInner() {
   // NOLINTNEXTLINE(cert-err58-cpp)
-  static std::unique_ptr<Logger> logger = std::unique_ptr<Logger>(new Logger());
+  static Logger* logger = new Logger();
```
blackyblack, line 13: Memory leak

PR #919, opened 2022-01-09
Add Btc BlockHeader, Block, Transaction structs

Mr-Leshiy:


include/veriblock/abfi/bitcoin/block.hpp
```diff
@@ -0,0 +1,31 @@
+#include "veriblock/abfi/bitcoin/serialize.hpp"
+#include "veriblock/pop/entities/btcblock.hpp"
+
+namespace btc {
```
Warchant, line 4: add altintegration namespace everywhere

src/pop/write_stream.cpp
```diff
@@ -18,4 +19,8 @@ const WriteStream::storage_t &WriteStream::data() const noexcept {
   return m_data;
 }
 
+std::string WriteStream::HexData() const noexcept {
```
Warchant, line 22:
```suggestion
std::string WriteStream::hex() const noexcept {
```

PR #939, opened 2022-02-06
Remove signals from the mempool

Mr-Leshiy:
needed for this PR https://github.com/VeriBlock/vbk-ri-btc/pull/660


Warchant (changes requested):
I think we can keep signals

PR #947, opened 2022-02-17
Build C wrapper and go tests for coverage

Warchant:


Mr-Leshiy:
@Warchant you need also enable `-DTESTING=ON` if you want to enable go tests

PR #954, opened 2022-02-22
Reractor of the mock_miner, pop_test_fixture

Mr-Leshiy:


include/veriblock/pop/mock_miner.hpp
```diff
@@ -119,7 +119,12 @@ class MockMiner {
   const VbkChainParams& vbkParams() const { return vbk_params_; }
   const BtcChainParams& btcParams() const { return btc_params_; }
 
-  MockMiner() {
```
blackyblack, line 122: You could keep the default constructor and avoid some duplicate code.

src/pop/c/mock_miner.cpp
```diff
@@ -18,7 +18,7 @@ POP_ENTITY_FREE_SIGNATURE(mock_miner) {
   }
 }
 
-POP_ENTITY_NEW_FUNCTION(mock_miner) { return new POP_ENTITY_NAME(mock_miner); }
+POP_ENTITY_NEW_FUNCTION(mock_miner) { return nullptr; }
```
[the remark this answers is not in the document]
Mr-Leshiy: It was temporarily, fixed :)

include/veriblock/pop/mock_miner.hpp
Mr-Leshiy: I am not sure that is a good idea, the main purpose is it externally setup MockMiner, so if you want to init it with RegTest version you should do it explicitly

blackyblack (approved):

Warchant (approved):

Issue #961, opened 2022-03-07
Assertion failed at /opt/alt-integration-cpp-x/src/pop/base58.cpp:160 inside AssertDecodeBase58


Mr-Leshiy:
```
2022-03-07T06:22:42Z [alt-cpp] [debug] overrideTip: ALT:56714:e218cd87f640e053ff02e0e6a5dec32023f102614549b0d97c03000000000000 VBK:3007905:00000000022cf637ff6e344cecc5db2da830eff38d19a3ac BTC:726123:00000000000000000003e2b0974c08fe61c8d21846cfa4056c2390e331dda6e5
2022-03-07T06:22:42Z     - Verify 0 txins: 91.59ms (0.000ms/txin) [3.54s (35.39ms/blk)]
2022-03-07T06:22:42Z     - Index writing: 0.04ms [0.00s (0.04ms/blk)]
2022-03-07T06:22:42Z     - Callbacks: 0.01ms [0.00s (0.01ms/blk)]
2022-03-07T06:22:42Z   - Connect total: 91.70ms [3.93s (39.29ms/blk)]
2022-03-07T06:22:42Z   - Flush: 0.01ms [0.00s (0.01ms/blk)]
2022-03-07T06:22:42Z   - Writing chainstate: 0.01ms [0.00s (0.01ms/blk)]
2022-03-07T06:22:42Z Blockpolicy estimates updated by 0 of 0 block txs, since last block 0 of 0 tracked, mempool map size 0, max target 0 from current
2022-03-07T06:22:42Z [alt-cpp] [critical] AssertDecodeBase58: Assertion failed at /opt/alt-integration-cpp-1beffc9d12147caa221e2cc0c440ea7c1fd675a8/src/pop/base58.cpp:160 inside AssertDecodeBase58:
status
Invalid input: �h���\x7f\x00\x00JzYhAVCx6vYvvKnWoDFq4Q

Assertion failed at /opt/alt-integration-cpp-1beffc9d12147caa221e2cc0c440ea7c1fd675a8/src/pop/base58.cpp:160 inside AssertDecodeBase58:
status
Invalid input: �h���JzYhAVCx6vYvvKnWoDFq4Q
terminate called without an active exception
```

Mr-Leshiy:
Not enough info to figure out what was the problem and how reproduce it.
Added additional logging.

Warchant:
Closing because it is no longer reproducible.

PR #963, opened 2022-03-07
Update base58/base59 decoding error logging

Mr-Leshiy:
It is logging improvement to later catch https://github.com/VeriBlock/alt-integration-cpp/issues/961 issue

### DIFF
--- a/src/pop/entities/address.cpp
+++ b/src/pop/entities/address.cpp
@@ -50,9 +50,9 @@ static bool isBase58String(const std::string& input, ValidationState& state) {
   return DecodeBase58(input, out, state);
 }
 
-static bool isBase59String(const std::string& input) {
+static bool isBase59String(const std::string& input, ValidationState& state) {
   std::vector<uint8_t> out;
-  return DecodeBase59(input, out);
+  return DecodeBase59(input, out, state);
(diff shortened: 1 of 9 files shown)


src/pop/entities/address.cpp
```diff
@@ -45,14 +45,14 @@ static std::string getChecksumPortionFromAddress(const std::string& address,
   return address.substr(MULTISIG_ADDRESS_DATA_END + 1);
 }
 
-static bool isBase58String(const std::string& input) {
+static bool isBase58String(const std::string& input, ValidationState& state) {
```
Warchant, line 48: use dummy "state" instead of new argument

src/pop/entities/address.cpp
```diff
@@ -45,14 +45,14 @@ static std::string getChecksumPortionFromAddress(const std::string& address,
...
   std::vector<uint8_t> out;
-  return DecodeBase58(input, out);
+  return DecodeBase58(input, out, state);
 }
 
-static bool isBase59String(const std::string& input) {
+static bool isBase59String(const std::string& input, ValidationState& state) {
```
Warchant, line 53: also here

PR #973, opened 2022-03-25
Disable vbk, btc finalization

Mr-Leshiy:


Warchant:
Do not disable finalization

PR #975, opened 2022-03-30
Fix and enable altblock_finalization_test_2, altblock_finalization_test_3 

Mr-Leshiy:
Closes https://github.com/VeriBlock/alt-integration-cpp/issues/974


src/pop/blockchain/mempool_block_tree.cpp
```diff
@@ -42,7 +42,7 @@ bool MemPoolBlockTree::checkContextually(const VbkBlock& block,
 
   auto* tip = vbkstable.getBestChain().tip();
   VBK_ASSERT(tip);
-  bool tooOld = abs(tip->getHeight() - block.getHeight()) >
+  bool tooOld = tip->getHeight() - block.getHeight() >
```
Warchant, line 45: Why?

Mr-Leshiy: `tooOld` property is only for blocks which are below the tip

Warchant: I don't see the logic. Tip can be lower in case if it's POP score is higher than second chain, but second chain is higher.

Mr-Leshiy: We are applying different blocks into the `MemPoolBlockTree` and it is possible that we will apply some blocks which are higher/next blocks to the current tip, but will not apply it into the tree.

